PR #4, opened 2020-11-28
feat: adjustments for APISIX 2.1 release

gxthrj:
Changes:
1. Add `namespace` for Kinds.
2. Explicit configure `allow_admin` in  `configmap`, because the `allow_admin` in `config.default.yaml` is configured.
3. Adjust `etcd` default configuration.
4. Remove `plugins` field if there is no plugin list, so the plugin list in `config.default.yaml` will be used.
5. Add etcd dependency install by default.
6. Modify README .


moonming (changes requested):
why add etcd bin file to repo？

gxthrj:
> why add etcd bin file to repo？

This is the helm way for installing dependencies.

[We can find Postgresql in Kong Charts](https://github.com/Kong/charts/tree/main/charts/kong/charts)

chart/apisix/Chart.yaml
```diff
@@ -31,9 +31,16 @@ type: application
...
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 1.16.0
+appVersion: 2.1.0
+
+dependencies:
+  - name: etcd
+    version: 5.2.1
```
membphis, line 43: `5.2.1`? I am confused with this version value

chart/apisix/values.yaml
```diff
@@ -95,34 +102,51 @@ admin:
     admin: edd1c9f034335f136f87ad84b625c8f1
     viewer: 4054f7cf07e344346cd3f287985e76a2
 
+allow:
+  # The ip range for allowing access to Apache APISIX
+  ipList:
+    - 127.0.0.1/24
+
 # APISIX plugins to be enabled
 plugins:
```
membphis, line 111: I think we can delete `plugins`, use the default list should be enough

chart/apisix/values.yaml
```diff
@@ -95,34 +102,51 @@ admin:
...
   - udp-logger
+  - uri-blocker
   - wolf-rbac
-  - proxy-cache
-  - tcp-logger
-  - proxy-mirror
-  - kafka-logger
-  - cors
+  - zipkin
 stream_plugins:
```
membphis, line 150: ditto

chart/apisix/Chart.yaml
gxthrj: This is etcd Charts version ，not the version of etcd.
It is Chart dependencies here.

gxthrj: In Charts, version is for Charts, appVersion is for the application.

chart/apisix/values.yaml
gxthrj: I keep it here is to leave a configuration item(placeholder) for users who wanted to use plugins belong themself.
 In fact, we can indeed set the plug-in list to be empty, I made it compatible.
But I am not sure if we should set the plugin list empty by default?

chart/apisix/values.yaml
gxthrj: I keep it here is to leave a configuration item(placeholder) for users who wanted to use plugins belong themself.
In fact, we can indeed set the plug-in list to be empty, I made it compatible.
But I am not sure if we should set the plugin list empty by default?

chart/apisix/Chart.yaml
```diff
@@ -31,9 +31,16 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.0
+version: 2.1.0
```
ShiningRush, line 34: chart version is different with application version, our chart is not validated by many production, it is not stable now, so `0.x.x` would be better. `2.1.1` tells people it already production ready.

chart/apisix/values.yaml
membphis: only comment should be enough if you want the user to know how to use it

chart/apisix/templates/configmap.yaml
```diff
@@ -72,9 +73,14 @@ data:
       #    disk_path: "/tmp/disk_cache_two"
       #    cache_levels: "1:2"
 
-    #  allow_admin:                  # http://nginx.org/en/docs/http/ngx_http_access_module.html#allow
-    #    - 127.0.0.0/24              # If we don't set any IP list, then any IP access is allowed by default.
-    #    - 172.17.0.0/24
+      allow_admin:                  # http://nginx.org/en/docs/http/ngx_http_access_module.html#allow
```
ShiningRush, line 76: It is not a good design, we can not limit user config by predefined parameter.
Define a values parameter names `config`, then transform it's content to `config.yaml` would be better, it can give user full control to config.

lilien1010:
how about remove etcd binary file and use dependencies？

chart/apisix/templates/tests/test-connection.yaml
```diff
@@ -18,6 +18,7 @@ apiVersion: v1
 kind: Pod
 metadata:
   name: "{{ include "apisix.fullname" . }}-test-connection"
+  namespace: {{ .Release.Namespace }}
```
ShiningRush, line 21: We really need this pod? I think the health check is enough.

chart/apisix/templates/service-gateway.yaml
```diff
@@ -18,6 +18,7 @@ apiVersion: v1
 kind: Service
```
ShiningRush, line 18: Can this service work well?I found that there are some problems:
- it not config `selector`, it will not bind to any pods
- no ports specified, it will not proxy any traffic

chart/apisix/templates/ingress.yaml
```diff
@@ -25,6 +25,7 @@ apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
   name: {{ $fullName }}
+  namespace: {{ .Release.Namespace }}
```
ShiningRush, line 28: `Ingress` required cluster install `ingress-controller`, k8s cluster does not install any implement about it by default, so this CRD will not work.

chart/apisix/Chart.yaml
gxthrj: I am going to use the third minor version number as the helm version, keeping the first two consistent with APISIX. 
e.g. 2.1.x for chart version.

gxthrj: Ok, use 0.x.x first

chart/apisix/templates/configmap.yaml
gxthrj: Okay, we can do this config.yaml configuration in another issue #7 .

chart/apisix/templates/ingress.yaml
gxthrj: We can remove this first, and add it back when apisix-ingress-controller is ready.

chart/apisix/templates/service-gateway.yaml
gxthrj: `selector` is in the end of this file. It is defined in `_helpers.tpl`
`ports` is defined, the value is `gateway` in values.yaml

ShiningRush: Got it, hmmm, it is really too long

gxthrj:
> how about remove etcd binary file and use dependencies？

Ok, Apache also does not allow binary files in the project.
I will remove the etcd binary and add some command in README for dependeny installation.

ShiningRush (approved):

chart/apisix/templates/configmap.yaml
```diff
@@ -137,12 +143,19 @@ data:
         #  ipc_shared_dict: 100m        # custom shared cache, format: `cache-key: cache-size`
 
     etcd:
+    {{- if .Values.etcd.enabled }}
+      host:                                 # it's possible to define multiple etcd hosts addresses of the same etcd cluster.
+        - "http://{{ .Release.Name }}-etcd.{{ .Release.Namespace }}.svc.cluster.local:2379"
```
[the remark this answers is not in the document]
gxthrj: Done

chart/apisix/templates/deployment.yaml
```diff
@@ -77,13 +78,6 @@ spec:
               subPath: config.yaml
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
```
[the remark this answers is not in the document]
gxthrj: Done

lilien1010 (approved):
LGTM

moonming (approved):

moonming (approved):

PR #19, opened 2021-01-13
feat: support to expose admin api

tokers:


chart/apisix/templates/configmap.yaml
```diff
@@ -82,7 +82,9 @@ data:
         - 0.0.0.0/0
       {{- end}}
       #   - "::/64"
-      # port_admin: 9180              # use a separate port
+      {{- if .Values.admin.enabled }}
+      port_admin: {{ .Values.admin.containerPort }}
```
totemofwolf, line 86: Why not use `admin.port`

lilien1010: I used to call it admin_port

lilien1010 (approved):

chart/apisix/templates/configmap.yaml
tokers: I think `.Values.admin.admin_port` is redundant, maybe `.Values.admin.port` is better.

tokers: @totemofwolf @lilien1010 Changed.

membphis (approved):

chart/apisix/values.yaml
```diff
@@ -97,6 +97,15 @@ dns:
 admin:
   # Enable Admin API
   enabled: true
+  # admin service type
+  type: ClusterIP
+  # loadBalancerIP: a.b.c.d
+  # loadBalancerSourceRanges:
+  #   - "143.231.0.0/16"
```
shreemaan-abhishek, line 104: @membphis why are these comments needed? 🤔

PR #71, opened 2021-05-09
chore: update the default version of apisix-dashboard

Firstsawyou:
Currently, the default installation version of APISIX is `2.5`, but the default installation version of apisix-dashboard is `2.3`, and the versions between them are not compatible. This PR is to update the default installed version of apisix-dashboard to 2.6, which solves the problem of incompatible versions when apisix-dashboard and apisix are installed by default.

https://github.com/apache/apisix-helm-chart/blob/c7edd2da0bac6a1c350478226cbee3c20a9556bc/charts/apisix/values.yaml#L22

https://github.com/apache/apisix-helm-chart/blob/c7edd2da0bac6a1c350478226cbee3c20a9556bc/charts/apisix-dashboard/values.yaml#L27

### DIFF
--- a/charts/apisix-dashboard/values.yaml
+++ b/charts/apisix-dashboard/values.yaml
@@ -24,7 +24,7 @@ image:
   repository: apache/apisix-dashboard
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
-  tag: "2.3"
+  tag: "2.6"
 
 imagePullSecrets: []
 nameOverride: ""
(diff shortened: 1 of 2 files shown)


charts/apisix-dashboard/values.yaml
```diff
@@ -24,7 +24,7 @@ image:
   repository: apache/apisix-dashboard
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
-  tag: "2.3"
+  tag: "2.6"
```
tokers, line 27: Also update the verison of chart in Chart.yaml.

Firstsawyou: Updated.

Firstsawyou:
@gxthrj  Please take a look at this PR. Some users have reported this problem. This PR needs to be merged as soon as possible.

tao12345666333 (approved):

gxthrj (approved):

PR #73, opened 2021-05-10
feat: support for controller-only deployment

tao12345666333:
Add new top-level `apisix.enabled` setting to control whether
deploy apisix-ingress controller only.

Signed-off-by: Jintao Zhang <zhangjintao9020@gmail.com>

### DIFF
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -14,6 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+apisix:
+  # Enable or disable Apache APISIX itself
+  # Set it to flase and ingress-controller.enabled=true will deploy only ingress-controller
+  enabled: true
+
 replicaCount: 1
 image:
   repository: apache/apisix


charts/apisix/values.yaml
```diff
@@ -14,6 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+apisix:
+  # Enable or disable Apache APISIX itself
+  # Set it to flase and ingress-controller.enabled=true will deploy only ingress-controller
+  enabled: true
```
tokers, line 20: There are so many top-level fields which are related with APISIX, I think we can put all of them under the apisix field.

tokers (approved):

charts/apisix/values.yaml
tao12345666333: Yep. I did not modify it because I was worried about breaking changes.

gxthrj (approved):

charts/apisix/values.yaml
gxthrj: There is no place to reference this value yet

PR #89, opened 2021-06-03
chore: add externalTrafficPolicy for apisix gateway service

### LINKED ISSUE #88: Apisix Gateway Ingress ```externalTrafficPolicy``` missing setting
Currently, you are unable to change ```externalTrafficPolicy``` via the apisix helm chart for the [service gateway template](https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/templates/service-gateway.yaml).

This is useful if you want to change from ```Cluster``` to ```Local```

tao12345666333:
Signed-off-by: Jintao Zhang <zhangjintao9020@gmail.com>

closes #88

### DIFF
--- a/charts/apisix/templates/service-gateway.yaml
+++ b/charts/apisix/templates/service-gateway.yaml
@@ -28,6 +28,7 @@ metadata:
     {{- include "apisix.labels" . | nindent 4 }}
 spec:
   type: {{ .Values.gateway.type }}
+  externalTrafficPolicy: {{ .Values.gateway.externalTrafficPolicy }}
   {{- if eq .Values.gateway.type "LoadBalancer" }}
   {{- if .Values.gateway.loadBalancerIP }}
   loadBalancerIP: {{ .Values.gateway.loadBalancerIP }}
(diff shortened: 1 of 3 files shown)


charts/apisix/templates/service-gateway.yaml
```diff
@@ -28,6 +28,7 @@ metadata:
     {{- include "apisix.labels" . | nindent 4 }}
 spec:
   type: {{ .Values.gateway.type }}
+  externalTrafficPolicy: {{ .Values.gateway.externalTrafficPolicy }}
```
tokers, line 31: We can set this field only if the service type is `LoadBalancer`.

tao12345666333: Both `LoadBalancer` and `NodePort` are affected by this parameter.

tao12345666333: Both `LoadBalancer` and `NodePort` are affected by this parameter. 
I don’t think it is necessary to add judgment here.

gxthrj (approved):

charts/apisix/templates/service-gateway.yaml
tokers: Got it.

tokers (approved):

PR #106, opened 2021-06-23
Update etcd and dashboard

caryyu:
In order to minimize the ability on the initial setup, there're still a few configs should be refined, which are:

1. the `etcd` statefulset replica number
2. the `etcd` persistence instruction - a direct reference to mitigate the usage
3. ~~the `etcd` endpoints for `dashboard` - which sometimes, the different helm release name could make an impact on it~~

### DIFF
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -214,6 +214,10 @@ etcd:
   statefulset:
     replicaCount: 3
 
+  persistence:
+    enabled: true
+    size: 8Gi
+    # storageClass: "-"
 
 dashboard:
   enabled: false
@@ -222,8 +226,6 @@ dashboard:
       etcd:
         endpoints:
           - apisix-etcd:2379
-        username: admin
-        password: edd1c9f034335f136f87ad84b625c8f1
 
 ingress-controller:
   enabled: false
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -219,8 +219,10 @@ etcd:
     size: 8Gi
     # storageClass: "-"
 
+
 dashboard:
   enabled: false
+
 
 ingress-controller:
   enabled: false
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -211,12 +211,19 @@ etcd:
   service:
     port: 2379
 
-  replicaCount: 3
+  statefulset:
+    replicaCount: 3
 
 
 dashboard:
   enabled: false
-
+  config:
+    conf:
+      etcd:
+        endpoints:
+          - apisix-etcd:2379
+        username: admin
+        password: edd1c9f034335f136f87ad84b625c8f1
 
 ingress-controller:
   enabled: false
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -221,11 +221,6 @@ etcd:
 
 dashboard:
   enabled: false
-  config:
-    conf:
-      etcd:
-        endpoints:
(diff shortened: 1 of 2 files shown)


gxthrj:
cc @tao12345666333

charts/apisix/values.yaml
```diff
@@ -211,12 +211,19 @@ etcd:
...
 dashboard:
   enabled: false
-
+  config:
+    conf:
+      etcd:
+        endpoints:
+          - apisix-etcd:2379
+        username: admin
+        password: edd1c9f034335f136f87ad84b625c8f1
```
tokers, line 226: is this the fixed password in bitnami/etcd?

caryyu: Oops, misunderstood the integral part of `dashboard api server`, which I thought it should've had a credential or something, let me remove it and change something else, thanks for pointing out

tao12345666333:
thanks，let me take a look.

caryyu:
@tokers Please take a look again, hope it makes a lot sense to you guys :)

tokers (approved):

charts/apisix/values.yaml
```diff
@@ -211,12 +211,21 @@ etcd:
...
 
+  persistence:
+    enabled: true
+    size: 8Gi
+    # storageClass: "-"
 
 dashboard:
   enabled: false
-
+  config:
```
gxthrj, line 224: If we add field in values, Please show us where it will be used.
I think we can not add dashboard configuration here, but should add it into `chars/apsix-dashboard/values.yaml`.
cc @tao12345666333 Am I right ?

gxthrj (changes requested):

charts/apisix/values.yaml
caryyu: @gxthrj Yeah, I'm very aware of this DRY(dont-repeat-yourself) stuff, that's why I explained in the description above that a minimized config for `dashboard` is necessary due to the different helm release name would cause a potential issue, which the service name of `etcd` would have another name that would make `dashboard` failing to connect, however, I still partially agree with you as it is a nice-to-have stuff

tao12345666333: We currently have two main usage scenarios

* `helm install apisix apisix/apisix --set dashboard.enabled=true`  install apisix and apisix-dashboard.
* `helm install dashboard apisix/apisix-dashboard` only install apisix-dashboard.

In fact, here we use the subchart method for quoting, just make sure that the configuration here is consistent with the configuration in subchart.

But as we hope, we hope to reduce repeat works.
Let me have a quick try and give you an answer later

caryyu: Sure thing, let's do what you advised

caryyu:
@tao12345666333 Please take a review again, thanks.

charts/apisix/values.yaml
```diff
@@ -211,7 +211,13 @@ etcd:
   service:
     port: 2379
 
-  replicaCount: 3
+  statefulset:
+    replicaCount: 3
```
tao12345666333, line 215: We need to make the configuration here consistent with bitnami/etcd https://github.com/bitnami/charts/blob/080ab2485dfe9ba060710aad18c2edb6381fa543/bitnami/etcd/values.yaml#L249-L251

tao12345666333:
Thanks for the contribution, but it looks like we don't need it at the moment. The configuration of the etcd part, I personally still recommend keeping it consistent with bitnami/etcd.
I will close this one. If you have any question, please feel free to reopen it, or add comments

PR #110, opened 2021-07-02
Add a complete example of installation by helm

Hugo-X:
- support expose dashboard with nodePort 
- add a complete example


tao12345666333 (commented):
Thanks for your contributions!

We can now complete all installations directly with one command.

`helm install apisix charts/apisix --set ingress-controller.enabled=true   --set dashboard.enabled=true --set etcd.persistence.size="1Gi" --namespace ingress-apisix`

We don’t need to use the apisix-ingress-controller or dashboard chart separately for installation, unless we use the existing Apache APISIX cluster.

PR #114, opened 2021-07-13
🔧  Add json schema to validate helm3 values.

CloudXiaobai:
Helm v3 added support to validate values in a chart's values.yaml file with JSON schemas.the json schema can also be  provide a form for other webapp

### DIFF
--- a/charts/apisix/values.schema.json
+++ b/charts/apisix/values.schema.json
@@ -1,0 +1,255 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "type": "object",
+  "properties": {
+    "apisix": {
+      "type": "object",
+      "title": "APISIX Setting(s)",
+      "form": true,
+      "properties": {
+        "enabled": {
+          "type": "boolean",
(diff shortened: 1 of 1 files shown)


tokers:
@CloudXiaobai It seems that this schema is not complete.

CloudXiaobai:
> @CloudXiaobai It seems that this schema is not complete.

Yes, I think it is sufficient to check only the core parts, most users will not bother to configure the full value file.
Of course,  if necessary, we can completed the values

tokers:
> Yes, I think it is sufficient to check only the core parts, most users will not bother to configure the full value file.

We may use some tools to help us to generate such a jsonschema file.

tokers:
> > @CloudXiaobai It seems that this schema is not complete.
> 
> Yes, I think it is sufficient to check only the core parts, most users will not bother to configure the full value file.
> Of course, if necessary, we can completed the values

I know it would be tedious to give a complete description, but we should try to cover all of them. We may split this task into several PRs, in each PR, we can cover a section there.

PR #127, opened 2021-08-10
feat: add  stream_proxy only field

gxthrj:


charts/apisix/Chart.yaml
```diff
@@ -49,7 +49,7 @@ dependencies:
     condition: dashboard.enabled
     alias: dashboard
   - name: apisix-ingress-controller
-    version: 0.5.0
+    version: 0.6.0
```
tao12345666333, line 52: When modifying the dependency, you need to execute the following command
```
helm dependency update charts/apisix
```

tao12345666333 (changes requested):

charts/apisix/Chart.yaml
tao12345666333: I have updated it

tao12345666333 (approved):

tokers (approved):

PR #157, opened 2021-10-13
docs: Remove the "$" or "➜" symbol at the beginning of the command

nic-6443:
Remove the "$" or "➜" symbol at the beginning of the command, otherwise the command cannot be executed directly.


nic-6443:
> Don’t merge yet, I found some commands start with "➜", I solved them all at once.

All the places are fixed and now ready to be merged.

PR #175, opened 2021-11-03
feat: support more APISIX features

tokers:
This Pull Request adds some new features for the apisix charts, so that more and more APISIX features can be used.

1. Support to configure ETCD SNI;
2. Support to specify Plugin Attrs;
3. Support to use Pod UID as the APISIX instance id (see https://github.com/apache/apisix/issues/5417 for details).

Signed-off-by: Chao Zhang <tokers@apache.org>

### DIFF
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -186,6 +186,8 @@ stream_plugins:
   - ip-restriction
   - limit-conn
 
+pluginAttrs: null
+
 # customPlugins allows you to mount your own HTTP plugins.
 customPlugins:
   enabled: false
@@ -265,6 +267,7 @@ etcd:
       certFilename: ""
       certKeyFilename: ""
       verify: true
+      sni: ""
 
   service:
     port: 2379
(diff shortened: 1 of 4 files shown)


charts/apisix/values.yaml
```diff
@@ -186,6 +188,8 @@ stream_plugins:
   - ip-restriction
   - limit-conn
 
+pluginAttrs: null
```
tao12345666333, line 191: Its type should be array, right?

tokers: It's a map, let me to change to `{}`.

lingsamuel (approved):

tao12345666333 (approved):

tao12345666333:
Please also update the document to introduce these configuration items, and then we can merge.

tokers:
> Please also update the document to introduce these configuration items, and then we can merge.

Added.

tao12345666333 (approved):
LGTM

PR #183, opened 2021-11-26
feat(apisix): add service monitor support

### LINKED ISSUE #182: feat: add service monitor for main chart
We already have service monitor for ingress controller (#76), shall we make it available also for main chart?

webup:
Resolves #182 as a follow-up enhancement per this hands-on tutorial: https://mp.weixin.qq.com/s/MB6ftEii5OCpicMm872zYg.

### DIFF
--- a/charts/apisix/templates/configmap.yaml
+++ b/charts/apisix/templates/configmap.yaml
@@ -226,6 +226,13 @@ data:
 
     {{- if .Values.customPlugins.enabled }}
     plugin_attr:
+      {{- if .Values.serviceMonitor.enabled }}
+      prometheus:
+        export_addr:
+          ip: 0.0.0.0
+          port: {{ .Values.serviceMonitor.containerPort }}
+        export_uri: {{ .Values.serviceMonitor.path }}
+      {{- end }}
     {{- range $plugin := .Values.customPlugins.plugins }}
       {{ $plugin.name }}: {{- $plugin.attrs | toYaml | nindent 8 }}
     {{- end }}
--- a/charts/apisix/templates/servicemonitor.yaml
+++ b/charts/apisix/templates/servicemonitor.yaml
@@ -1,0 +1,39 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


charts/apisix/templates/configmap.yaml
```diff
@@ -226,6 +226,13 @@ data:
 
     {{- if .Values.customPlugins.enabled }}
     plugin_attr:
+      {{- if .Values.serviceMonitor.enabled }}
+      prometheus:
+        export_addr:
+          ip: 0.0.0.0
```
tokers, line 232: It's dangerous and if APISIX instances are exposed to the Internet, the metrics data can be fetched by others.

webup: So how shall we expose metrics from APISIX gateway, any better approach?

tokers: Maybe we can use the PodMonitor to fetch metrics, then we don't have to export the port to the service, and at least accessors outside the cluster cannot snoop the metrics.

webup: Got your point, while if you check the codes, will find that no ports were exposed in `Service`, we only expose one essential container port in `Deployment`, and in this way `ServiceMonitor` could still work well.

This [issue](https://github.com/prometheus-operator/prometheus-operator/issues/3119#issuecomment-631409690) may echo my understanding: "It will scrape all pods behind the service, because the `Service` maintains an `Endpoints` object. That's where Prometheus discovers the targets from."

tokers: Is the endpoint port defined on ServiceMonitor is a container port? If so, I think is OK.

webup: Yes, that's container port, not service port. :)

Gallardot:
LGTM

I tested this feature in my environment

gwAl7sqJyV

charts/apisix/templates/configmap.yaml
tokers: OK.

charts/apisix/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - {{ .Values.serviceMonitor.namespace | default .Release.Namespace }}
+  selector:
+    matchLabels:
+      {{- include "apisix.labels" . | nindent 6 }}
+  endpoints:
+  - scheme: http
+    targetPort: prometheus
+    path: {{ .Values.serviceMonitor.path }}
+    interval: {{ .Values.serviceMonitor.interval }}
+{{- end }}
```
tokers, line 39: Let's add an EOL symbol.

Gallardot:
@webup 
Thanks for your contribution! Great job!

Similar to #185 , we want this feature to also add extra labels and annotations. What do you think?

tao12345666333:
@webup Thanks! Can you resolve the conflicting?

tao12345666333:
Using #192 instead. Thanks! I will close this one.

PR #87, opened 2021-05-31
feat: add service monitor

tao12345666333:
Signed-off-by: Jintao Zhang <zhangjintao9020@gmail.com>

fixes: #76

Issue #76, opened 2021-05-17
Observability support 


tao12345666333:
we should add `ServiceMonitor` resource for observability.

tao12345666333:
please assign to me. @tokers  @gxthrj

gxthrj:
> we should add `ServiceMonitor` resource for observability.

agree +1

PR #185, opened 2021-12-02
feat: add extra labels and annotations for service monitor

Gallardot:
If [serviceMonitorSelector.matchLabels](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/getting-started.md#include-servicemonitors) are enabled, we need some extra labels for service monitor

link: #76 #87

### DIFF
--- a/charts/apisix-ingress-controller/values.yaml
+++ b/charts/apisix-ingress-controller/values.yaml
@@ -124,3 +124,7 @@ serviceMonitor:
   enabled: false
   namespace: "monitoring"
   interval: 15s
+  # @param serviceMonitor.labels ServiceMonitor extra labels
+  labels: {}
+  # @param serviceMonitor.annotations ServiceMonitor annotations
+  annotations: {}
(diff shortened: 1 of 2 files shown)


charts/apisix-ingress-controller/values.yaml
```diff
@@ -124,3 +124,9 @@ serviceMonitor:
   enabled: false
   namespace: "monitoring"
   interval: 15s
+  ## @param serviceMonitor.labels ServiceMonitor extra labels
```
tao12345666333, line 127: You can delete the extra # to keep it consistent with other comments.
```suggestion
  # @param serviceMonitor.labels ServiceMonitor extra labels
```

tao12345666333 (approved):

charts/apisix-ingress-controller/values.yaml
Gallardot: done!

tokers (approved):

PR #196, opened 2021-12-15
feat(apisix): add service label

Gallardot:
The `servicemonitor` matches both `apisix-admin` and `apisix-gateway` services, and Prometheus generates two scrape jobs. Adding a `service` label would be more appropriate.

### DIFF
--- a/charts/apisix/templates/service-admin.yaml
+++ b/charts/apisix/templates/service-admin.yaml
@@ -25,6 +25,7 @@ metadata:
     {{- end }}
   labels:
     {{- include "apisix.labels" . | nindent 4 }}
+    app.kubernetes.io/service: apisix-admin
 spec:
   type: {{ .Values.admin.type }}
   {{- if eq .Values.admin.type "LoadBalancer" }}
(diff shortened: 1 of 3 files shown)


charts/apisix/templates/service-admin.yaml
```diff
@@ -25,6 +25,7 @@ metadata:
     {{- end }}
   labels:
     {{- include "apisix.labels" . | nindent 4 }}
+    app.kubernetes.io/component: apisix-admin
```
tao12345666333, line 28: In fact, from the semantics of `component`, they are the same componet.
I don’t think this label is reasonable enough.

Gallardot: How about `app.kubernetes.io/service` ?

tao12345666333: 👌

tao12345666333 (approved):
Thanks!

juzhiyuan (approved):
Thanks!

PR #199, opened 2021-12-21
feat: add networkpolicy to apisix helm chart

0xgui:
Add the ability to create a custom `networkpolicy` for APISIX helm chart.
The etcd helm chart already provides a `networkpolicy` if needed.
If this proves to be a good addition, I can also make the `networkpolicy` for the APISIX-dashboard and ingress controller helm chart.

### DIFF
--- a/charts/apisix/templates/networkpolicy.yaml
+++ b/charts/apisix/templates/networkpolicy.yaml
@@ -1,0 +1,77 @@
+{{- if .Values.apisix.networkPolicy.enabled }}
+{{- $fullName := include "apisix.fullname" . -}}
+kind: NetworkPolicy
+apiVersion: {{ template "networkPolicy.apiVersion" . }}
+metadata:
+  name: {{ include "apisix.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels: 
+    {{- include "apisix.labels" . | nindent 4 }}
+  {{- if .Values.apisix.networkPolicy.annotations }}
+    {{- with .Values.apisix.networkPolicy.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- end }}
+spec:
+  podSelector:
+    matchLabels: {{- include "apisix.selectorLabels" . | nindent 6 }}
+  policyTypes:
+    - Ingress
+  {{- if .Values.apisix.networkPolicy.additionalEgress }}
+    - Egress
+  egress:
+    - ports:
+        - port: 53
+          protocol: UDP
+    - ports:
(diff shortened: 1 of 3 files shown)


charts/apisix/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,77 @@
...
+              {{ $key | quote }}: {{ $value | quote }}
+              {{- end }}
+          {{- end }}
+        {{- end }}
+      {{- end }}
+    {{- if .Values.apisix.networkPolicy.additionalIngress }}
+      {{- with .Values.apisix.networkPolicy.additionalIngress }}
+        {{- toYaml . | nindent 4 }}
+      {{- end }}    {{- end }}
+{{- end }}
```
tokers, line 77: Add a EOS symbol.

charts/apisix/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,77 @@
...
+    matchLabels: {{- include "apisix.selectorLabels" . | nindent 6 }}
+  policyTypes:
+    - Ingress
+  {{- if .Values.apisix.networkPolicy.additionalEgress }}
+    - Egress
+  egress:
+    - ports:
+        - port: 53
+          protocol: UDP
+    - ports:
```
tokers, line 27: The egress rules are used to limit the destinations that APISIX can reach, so here use APISIX ports as the "target port" is pathological for me.

0xgui: Yes, sorry my bad. It was clearly a mistake. 
What about now ? Is the `egress` policy OK for you ?

tokers: I think we don't have to restrict the source port in the egress rules.

0xgui: So, should we remove all the egress port definition and let the user choose whatever ports in the egress they might want ?

tokers: I think specify the source port is meaningless, as the source ports usually specified by the kernel, it's somewhat random for users.

charts/apisix/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,72 @@
...
+    {{- end }}
+  {{- end }}
+spec:
+  podSelector:
+    matchLabels: {{- include "apisix.selectorLabels" . | nindent 6 }}
+  policyTypes:
+    - Ingress
+  {{- if .Values.apisix.networkPolicy.additionalEgress }}
+    - Egress
+  egress:
```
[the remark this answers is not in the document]
tokers: Cannot capture the contexts about this topic, could you give me some hints or links about it?

PR #201, opened 2021-12-26
feat: bump APISIX Dashboard to 2.10.1

bzp2010:


charts/apisix-dashboard/Chart.yaml
```diff
@@ -31,12 +31,12 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.3.0
+version: 0.4.1
```
tao12345666333, line 34: why using `0.4.1` version ?

bzp2010: First, we released version `2.10` of the dashboard, but it had some issues, so we went further and released a patch version of `2.10.1`. We skipped the `2.10` binary release, so I think it's a little better to skip the `0.4.0` version code. What do you think?

tokers: It seems that we don't have a mapping relationship between the version of apisix-dashboard and its helm chart.

tao12345666333: I think it is more appropriate to release 0.4.0. Because Helm chart does not have a direct strong mapping relationship with APISIX Dashboard.

bzp2010: OK, I will change it.

starsz (approved):

tokers (approved):

nic-chen (approved):

tao12345666333 (approved):

PR #232, opened 2022-02-11
feat(apisix): add timezone and extraEnvVars

Gallardot:
You may need to set the timezone to be consistent with your local time zone, otherwise the apisix's logs may used to retrieve event maybe in wrong timezone. 

Also define `extraEnvVars` to allow users to extend additional environment variables


charts/apisix/values.yaml
```diff
@@ -96,6 +96,16 @@ apisix:
   podAntiAffinity:
     enabled: false
 
+  # timezone is the timezone where apisix uses.
+  # For example: "UTC" or "Asia/Shanghai"
+  # This value will be set on apisix container's environment variable TZ.
+  # You may need to set the timezone to be consistent with your local time zone,
+  # otherwise the apisix's logs may used to retrieve event maybe in wrong timezone.
+  timezone: UTC
```
tokers, line 104: Should add the documents to explain it.

charts/apisix/values.yaml
```diff
@@ -96,6 +96,16 @@ apisix:
...
+
+  # extraEnvVars An array to add extra env vars
+  extraEnvVars: []
```
tokers, line 107: The format for the key and value should be explained.

charts/apisix/values.yaml
Gallardot: > Should add the documents to explain it.

I also add this note on the `README` document.  I think this explanation document about `timezone` might be more appropriate in the [APISIX FAQ](https://apisix.apache.org/docs/apisix/FAQ).   @tokers  WDYT？

charts/apisix/values.yaml
Gallardot: Done! PTAL.

tokers: Shall we also add it in the README?

charts/apisix/values.yaml
tokers: Agree with you, would you like to submit another PR to add it?

Gallardot: > Agree with you, would you like to submit another PR to add it?

OK，I will submit another PR first and then deal with the current PR

moonming:
the same as https://github.com/apache/apisix/issues/3039? we can not hard-code TZ

tokers:
@Gallardot We may do not set the `TZ` if the value is empty.

tao12345666333 (approved):
LGTM

Gallardot:
> @Gallardot We may do not set the `TZ` if the value is empty.

@tokers Done.   PTAL https://github.com/apache/apisix/pull/6705

charts/apisix/README.md
```diff
@@ -76,9 +76,10 @@ The following tables lists the configurable parameters of the apisix chart and t
...
 | `apisix.luaModuleHook.configMapRef.mounts[].key` | Name of the ConfigMap key, for setting the mapping relationship between ConfigMap key and the lua module code path. | `""` |
 | `apisix.luaModuleHook.configMapRef.mounts[].path` | Filepath of the plugin code, for setting the mapping relationship between ConfigMap key and the lua module code path. | `""` |
-| `apisix.timezone` | The timezone where APISIX container uses. For example: UTC, Asia/Shanghai. This value will be set on apisix container's environment variable TZ. Sometimes may need to set the timezone to be consistent with local time zone, otherwise the apisix's logs may used to retrieve event maybe in wrong timezone.                | `UTC`|
-| `apisix.extraEnvVars` | APISIX container extra environment variables                   | `[]`|
-
+| `apisix.timezone` | The timezone where APISIX container uses. For example: UTC, Asia/Shanghai. This value will be set on apisix container's environment variable TZ. Sometimes may need to set the timezone to be consistent with local time zone, otherwise the apisix's logs may used to retrieve event maybe in wrong timezone.                | nil|
```
tokers, line 79: The default value is not `nil`.

Gallardot: Done. PTAL.

tokers (approved):

PR #233, opened 2022-02-11
fix: enable stream_route mode when ingress enabled

tao12345666333:
Since the release of APISIX v2.12, we found that when using Helm chart to install both APISIX and Ingress controller, the Ingress controller does not work properly.

This is because the stream_route mode is disabled by default in APISIX, but it is required by the Ingress controller.

xref:
issue: https://github.com/apache/apisix-ingress-controller/issues/851#issuecomment-1026306420
https://github.com/apache/apisix-ingress-controller/issues/858

Another fix on the APISIX Ingress controller is tracked at https://github.com/apache/apisix-ingress-controller/issues/856


tao12345666333:
This is a serious problem, we need an upgrade version to release it immediately.

Gallardot:
Maybe we should update both Deployment and Service as well.

Gallardot:
> Why? This is for compatibility with Ingress controller behavior only, not for APISIX users. If the user really needs to expose stream_route, then modify the corresponding configuration.

Okay, got it

lingsamuel (approved):

bzp2010 (approved):
LGTM

PR #265, opened 2022-04-11
fix: ingress syntax support for Kubernetes 1.19 and later.

Senorsen:
Fix ingress syntax so that recent Kubernetes versions (>= 1.19) are supported.
Drop the support of Kubernetes 1.18 and lower version to keep it simple and clean.


Senorsen:
Seems duplicated

PR #268, opened 2022-04-19
nginx connection configurage to values.yaml attribute

PGDream:
with main snippet don`t modify nginx command `worker_rlimit_nofile、events.worker_connections、events.use`

### DIFF
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -121,6 +121,14 @@ apisix:
 nameOverride: ""
 fullnameOverride: ""
 
+serviceAccount:
+  create: false
+  annotations: {}
+  name: ""
+
+rbac:
+  create: false
+
 gateway:
   type: NodePort
   # If you want to keep the client source IP, you can set this to Local.
@@ -191,6 +199,7 @@ nginx:
   workerConnections: "10620"
   workerProcesses: auto
   enableCPUAffinity: true
+  envs: []
 
 # APISIX plugins to be enabled
 plugins:
@@ -288,15 +297,17 @@ extraVolumeMounts: []
 
 discovery:
   enabled: false
-  registry:
+  registry: {}
     # Integration service discovery registry. E.g eureka\dns\nacos\consul_kv
     # reference:
-    # https://apisix.apache.org/docs/apisix/discovery#configuration-for-eureka
-    # https://apisix.apache.org/docs/apisix/discovery/dns#service-discovery-via-dns
-    # https://apisix.apache.org/docs/apisix/discovery/consul_kv#configuration-for-consul-kv
-    # https://apisix.apache.org/docs/apisix/discovery/nacos#configuration-for-nacos
+    # https://apisix.apache.org/docs/apisix/discovery/#configuration-for-eureka
+    # https://apisix.apache.org/docs/apisix/discovery/dns/#service-discovery-via-dns
+    # https://apisix.apache.org/docs/apisix/discovery/consul_kv/#configuration-for-consul-kv
+    # https://apisix.apache.org/docs/apisix/discovery/nacos/#configuration-for-nacos
+    # https://apisix.apache.org/docs/apisix/discovery/kubernetes/#configuration
     #
     # an eureka example:
+    # ```
     # eureka:
     #   host:
     #     - "http://${username}:${password}@${eureka_host1}:${eureka_port1}"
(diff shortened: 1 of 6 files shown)


PGDream:
values.yaml add `events.use` purpose  `APACHE APISIX  ngx_tpl.lua` file override  `events.use` command on nginx.conf

tao12345666333:
plz check the CI errors

Gallardot (changes requested):
Maybe we also need to update `README.md`. Rest LGTM.

charts/apisix/values.yaml
```diff
@@ -312,6 +312,12 @@ logs:
   errorLog: "/dev/stderr"
   errorLogLevel: "warn"
 
+core:
+  worker_rlimit_nofile: 20480
+  events:
+    worker_connections: 10620
+    use: epoll
```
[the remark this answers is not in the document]
PGDream: > the apisix is a linux system, with `epoll` networker model  better handle `networker connection`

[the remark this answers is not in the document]
Gallardot: ref: [ngx_core_module](http://nginx.org/en/docs/ngx_core_module.html#use) and [events](http://nginx.org/en/docs/events.html) 

```
There is normally no need to specify it explicitly, because nginx will by default use the most efficient method.
```

So is it better to default to nil?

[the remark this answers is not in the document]
PGDream: yes， i have see, Now that the scene has been clarified, is the display clearer?

PGDream:
> plz check the CI errors

@tao12345666333  lost reviewer!

charts/apisix/values.yaml
```diff
@@ -312,6 +312,11 @@ logs:
   errorLog: "/dev/stderr"
   errorLogLevel: "warn"
 
+core:
```
tokers, line 315: Use `nginx` will be better, `core` is too generic.

tao12345666333: +1

PGDream: Is it possible `logs、core` node add parent node `nginx`.
`worker_rlimit_nofile` command ... on `nginx` core model.
example:
nginx:
   logs:
      enableAccessLog
   core:
      worker_rlimit_nofile:

Gallardot: > Is it possible `logs、core` node add parent node `nginx`. `worker_rlimit_nofile` command ... on `nginx` core model. example:
> 
> ```
> nginx:
>    logs:
>       enableAccessLog
>    core:
>       worker_rlimit_nofile: 
> ```

good idea.

PGDream: done

charts/apisix/values.yaml
```diff
@@ -303,14 +303,20 @@ discovery:
       #     send: 2000
       #     read: 5000
 
-# access log and error log configuration
-logs:
```
tokers, line 307: This is a broken change, I suggest not introducing it directly.

tao12345666333: +1

tao12345666333: @PGDream plz take a look

PGDream:
This is more convenient for later expansion and structuring. Is it considered possible to change this way?

tao12345666333:
@PGDream In order to maintain compatibility, we can support both configuration methods at the same time. After 3 versions, remove the old configuration.

PGDream:
> @PGDream In order to maintain compatibility, we can support both configuration methods at the same time. After 3 versions, remove the old configuration.

good idea

tokers (approved):

PGDream:
> @PGDream In order to maintain compatibility, we can support both configuration methods at the same time. After 3 versions, remove the old configuration.

done

charts/apisix/values.yaml
```diff
@@ -312,6 +312,21 @@ logs:
...
   errorLogLevel: "warn"
 
+# access and error log configuration
+nginx:
+  core:
+    worker_rlimit_nofile: 20480
+    events:
+      worker_connections: 10620
+  logs:
+    enableAccessLog: true
```
Gallardot, line 322: How does those `logs` configuration item work in the current version? Maybe need some compatibility code?

tao12345666333:
It's on my queue, I will do the final review today.

charts/apisix/README.md
```diff
@@ -201,6 +201,20 @@ discovery:
 | `logs.errorLog`                  | Error log path | `/dev/stderr` |
 | `logs.errorLogLevel`                  | Error log level | `warn` |
```
tokers, line 202: Mark these parameters as deprecated. We can remove them after a few releases.

charts/apisix/values.yaml
tokers: I think we need to change all the places that we use the log parameters.

PGDream:
@tokers where not change of log parameters?

tokers:
> @tokers where not change of log parameters?

FYI: https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/templates/configmap.yaml#L166

PGDream:
> > @tokers where not change of log parameters?
> 
> FYI: https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/templates/configmap.yaml#L166

modify more configure,account upgrade may error..

tokers:
> 

No. If you don't modify them, what's the point of introducing the `nginx_config.logs`?

PGDream:
> @PGDream In order to maintain compatibility, we can support both configuration methods at the same time. After 3 versions, remove the old configuration.



> > 
> 
> No. If you don't modify them, what's the point of introducing the `nginx_config.logs`?

@PGDream In order to maintain compatibility, we can support both configuration methods at the same time. After 3 versions, remove the old configuration. ?  

how to solve then above problems?

tao12345666333:
@PGDream What problem are you having now?

PGDream:
> In order to maintain compatibility, we can support both configuration methods at the same time. After 3 versions, remove the old configuration.

excuse me, how do it?

tokers:
> > In order to maintain compatibility, we can support both configuration methods at the same time. After 3 versions, remove the old configuration.
> 
> excuse me, how do it?

Let's add the `nginx.log` field but at the same time, don't change the [log parameters](https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/README.md#toc15).

PGDream:
> > > In order to maintain compatibility, we can support both configuration methods at the same time. After 3 versions, remove the old configuration.
> > 
> > 
> > excuse me, how do it?
> 
> Let's add the `nginx.log` field but at the same time, don't change the [log parameters](https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/README.md#toc15).

that`s it now

PR #283, opened 2022-05-12
apisix Deployment and DaemonSet securityContext reference error

zplinuxlover:
In apisix Deployment and DaemonSet, securityContext refer to .Values.podSecurityContext, But in values.yaml, there is no definition of it, I think this is a input mistake.

In apisix deployment.yaml

```
securityContext:
        {{- toYaml .Values.podSecurityContext | nindent 8 }}
```
should be

```
securityContext:
        {{- toYaml .Values.apisix.podSecurityContext | nindent 8 }}
```


tokers:
Thanks for the fix, I have run the CI jobs, let's see the result.

The changes LGTM.

PR #284, opened 2022-05-13
could not install apisix-ingress if ingress-controller.enabled changed to true when helm install apisix charts

zplinuxlover:
apisix chart and apisix-dashboard chart use {{ .Release.Namespace }}, when we change the config `ingress-controller.enabled` default behaviour, we could not install the apisix-ingress chart.

In apisix charts values.yaml
```
ingress-controller:
  enabled: false
```
changed to 
```
ingress-controller:
  enabled: true
```
I want to install apisix and apisix-ingress at the same time. but apisix-ingress could not install successfully. In apisix chart, we do not specify the namesapce, we use {{ .Release.Namespace }}, but in apisix-ingress charts, we use the configured namespace in values.yaml 

```
  apisix:
    serviceName: apisix-admin
    #serviceNamespace: ingress-apisix
```

the code in apisix-ingress-controller charts deployment.yaml, we use a different namespace configured in values.yaml, 

```
command: ['sh', '-c', "until nc -z {{ .Values.config.apisix.serviceName }}.{{ Values.config.apisix.serviceNamespace }}.svc.{{ .Values.clusterDomain }} {{ .Values.config.apisix.servicePort }} ;
```

the init container could not success, so the apisix ingress controller could not startup

image

we should use `.Release.Namespace` consistent with others charts, apisix charts and apisix-dashboard.

### DIFF
--- a/charts/apisix-ingress-controller/templates/configmap.yaml
+++ b/charts/apisix-ingress-controller/templates/configmap.yaml
@@ -55,7 +55,7 @@ data:
       apisix_route_version: {{ .Values.config.kubernetes.apisixRouteVersion | quote }}
       enable_gateway_api: {{ .Values.config.kubernetes.enableGatewayAPI }}
     apisix:
-      default_cluster_base_url: http://{{ .Values.config.apisix.serviceName }}.{{ .Values.config.apisix.serviceNamespace }}.svc.{{ .Values.clusterDomain }}:{{ .Values.config.apisix.servicePort }}/apisix/admin
+      default_cluster_base_url: http://{{ .Values.config.apisix.serviceName }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}:{{ .Values.config.apisix.servicePort }}/apisix/admin
       default_cluster_admin_key: {{ .Values.config.apisix.adminKey | quote }}
       default_cluster_name: {{ .Values.config.apisix.clusterName | quote }}
 kind: ConfigMap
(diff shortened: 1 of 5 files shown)


tao12345666333:
```
helm install apisix apisix/apisix --set gateway.type=NodePort --set ingress-controller.enabled=true --set ingress-controller.config.apisix.serviceNamespace=apisix --namespace apisix --create-namespace --set ingress-controller.config.apisix.serviceName=apisix-admin
```

Thanks, you can pass `--set ingress-controller.config.apisix.serviceNamespace=apisix` and ` --set ingress-controller.config.apisix.serviceName=apisix-admin` .

Release Namespace is not used directly because of the scenario of a standalone installation of the APISIX Ingress controller (although it is not recommended)

I don't think this change is reasonable.

tao12345666333 (commented):
```
helm install apisix apisix/apisix --set gateway.type=NodePort --set ingress-controller.enabled=true --set ingress-controller.config.apisix.serviceNamespace=apisix --namespace apisix --create-namespace --set ingress-controller.config.apisix.serviceName=apisix-admin
```

Thanks, you can pass `--set ingress-controller.config.apisix.serviceNamespace=apisix` and ` --set ingress-controller.config.apisix.serviceName=apisix-admin` .

Release Namespace is not used directly because of the scenario of a standalone installation of the APISIX Ingress controller (although it is not recommended)

I don't think this change is reasonable.

zplinuxlover:
should set serviceNamespace: ingress-apisix  from ingress-apisix to default, hence user can use `helm install apisix .` in apisix charts dir to install all the components including apisix, apisix-ingress, apisix-ingress-controller

tao12345666333:
But this won't work if APISIX and APISIX Ingress are in different namespaces

PR #285, opened 2022-05-13
[WIP]feat: Support auto-generate random `admin/view` key

### LINKED ISSUE #255: [feature]Generate random admin/view key when use `helm install`
helm template string function: [randAlphaNum, randAlpha, randNumeric, and randAscii](https://helm.sh/docs/chart_template_guide/function_list/#randalphanum-randalpha-randnumeric-and-randascii)

Auto generate password/admin-view key when that is default value. 

Print password when it successed by helm NOTES.txt https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix-dashboard/templates/NOTES.txt

https://github.com/apache/apisix-helm-chart/blob/c9a2679c76433ec5784aaeb910a8f8e1a360d52c/charts/apisix-dashboard/values.yaml#L76-L81

https://github.com/apache/apisix-helm-chart/blob/c9a2679c76433ec5784aaeb910a8f8e1a360d52c/charts/apisix/values.yaml#L158-L160

mangoGoForward:
Signed-off-by: mango <xu.weiKyrie@foxmail.com>

Fixed: #255

Use helm template string function: [randAlphaNum, randAlpha, randNumeric, and randAscii](https://helm.sh/docs/chart_template_guide/function_list/#randalphanum-randalpha-randnumeric-and-randascii) to generate password/admin-view key, and print it to helm NOTES.txt


charts/apisix-dashboard/values.yaml
```diff
@@ -76,6 +76,7 @@ config:
   authentication:
     secert: secert
     expireTime: 3600
+    userCreate: true
```
tokers, line 79: Better to update the README. Describe the meaning of this field, what's more, tell users that the Admin API key and APISIX Dashboard username, and password will be generated randomly the first time they install the release.

Also, a guide about how to show the credentials is suggested to write so that users know how can they see the credentials.

mangoGoForward: Done, and the step two in `NOTES.txt` will be tell users know how to get the credentials or account.

Gallardot: If I [install `apisix-ingress-controller`]( https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/values.yaml#L405) at the same time through subchart, how can I ensure that this [key](https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix-ingress-controller/values.yaml#L128) is consistent?

tokers: > 

I didn't capture the meaning? APISIX Ingress Controller doesn't use it.

tokers (approved):

charts/apisix-dashboard/values.yaml
mangoGoForward: > If I [install `apisix-ingress-controller`](https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/values.yaml#L405) at the same time through subchart, how can I ensure that this [key](https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix-ingress-controller/values.yaml#L128) is consistent?

I got your meaning and let me fix, thanks.

mangoGoForward: I have no good idea how to consistent the admin key between `apisix` and `apisix-ingress-controller`, because it is different when quote the variable twice, so we must get it from `configmap` when `apisix` installed.
Maybe we can tell users to install `apisix-ingress-controller` with `--set config.apisix.adminKey` filed which can execute the command `kubectl get cm {{ include "apisix.fullname" . }} -n {{ .Release.Namespace }} -o=jsonpath={.data} | sed -nr '/- name: "admin"/,+2 p' | awk -F": " '/key/{print $2}'` to get.

tokers: > I have no good idea how to consistent the admin key between apisix and apisix-ingress-controller, because it is different when quote the variable twice, so we must get it from configmap when apisix installed.
Maybe we can tell users to install apisix-ingress-controller with --set config.apisix.adminKey filed which can execute the command kubectl get cm {{ include "apisix.fullname" . }} -n {{ .Release.Namespace }} -o=jsonpath={.data} | sed -nr '/- name: "admin"/,+2 p' | awk -F": " '/key/{print $2}' to get.

But the case is we install APISIX and Ingress Controller **at the same time**.

tokers: @mangoGoForward It's my understanding that the admin key should be generated automatically on APISIX side, since APISIX is the Admin API provider while APISIX Ingress Controller is just a caller so the latter should refer to the position that the former saves the API key.

tao12345666333 (commented):
You can define the generation method in the helper and a variable to store the generated random number

The random number can be referenced elsewhere

charts/apisix-dashboard/values.yaml
mangoGoForward: Hi, @tokers @tao12345666333 
I've tried some ways to solve he admin key consistent between `APISIX` and `Ingress Controller`, but failed.
each time I execute it and will get a new value, so we must get it from `configmap` when `APISIX` installed, but this way cannot suite the case install APISIX and Ingress Controller at the same time.

PR #303, opened 2022-06-06
update apisix-ingress-controller crd to support hmacAuth

Fatpa:
Support hmacAuth for helm chart.
Needs to upgrade to the latest apisix-ingress-controller version, but I don't know which one now, I don't modify the values.yaml.

Ref to [#1035](https://github.com/apache/apisix-ingress-controller/pull/1035)

### DIFF
--- a/charts/apisix-ingress-controller/crds/customresourcedefinitions.yaml
+++ b/charts/apisix-ingress-controller/crds/customresourcedefinitions.yaml
@@ -140,6 +140,8 @@ spec:
                   - basicAuth
                 - required:
                   - keyAuth
+                - required:
+                  - hmacAuth
                 properties:
                   basicAuth:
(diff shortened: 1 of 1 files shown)


tao12345666333:
Sorry, we will not modify the CRD file until the new version (v1.5) of APISIX Ingress controller is released, because this change will not take effect in the current v1.4.1

Fatpa:
So we should wait for the v1.5 of Apisix Ingress controller and continue this PR?

tao12345666333:
Yep. when APISIX Ingress v1.5 released, we need update all CRDs and apisix-ingress-controller's image.

tao12345666333:
Thanks #331 has been merged, this one has been included.

I think we can close this one

PR #305, opened 2022-06-06
feat: bump APISIX's dashboard deps to 2.13.0

bzp2010:


charts/apisix/Chart.lock
```diff
@@ -4,9 +4,9 @@ dependencies:
   version: 7.0.4
 - name: apisix-dashboard
   repository: https://charts.apiseven.com
-  version: 0.5.0
+  version: 0.6.0
```
tokers, line 7: The current APISIX image is `2.14.1-alpine`, is it compatible with APISIX Dashboard `2.13.0`?

bzp2010: As far as I know, there are very few changes to core resources in APISIX 2.12-2.14, mainly the addition of plugins. Therefore version compatibility is mainly reflected in the fact that some plug-ins cannot be found, which is currently tolerable.
Also, there are even more plugins that use the current 2.11 difference, so it makes sense to update.

BTW, the dashboard compatibility work for 2.14 is in progress, and a new version will be released when it is completed.

tao12345666333: In addition, I think the compatibility of the new version of the dashboard should be better than the old version

tao12345666333 (approved):

juzhiyuan (approved):

guoqqqi (approved):

PR #310, opened 2022-06-29
Fix: when custom plugin's attr is null

fengxsong:
- fix when custom plugin's attributes is null, type of attributes is object
- make more nginx's configuration configurable in values.yaml

### DIFF
--- a/charts/apisix/templates/configmap.yaml
+++ b/charts/apisix/templates/configmap.yaml
@@ -165,14 +165,16 @@ data:
     nginx_config:                     # config for render the template to genarate nginx.conf
       error_log: "{{ .Values.logs.errorLog }}"
       error_log_level: "{{ .Values.logs.errorLogLevel }}"         # warn,error
-      worker_rlimit_nofile: 20480     # the number of files a worker process can open, should be larger than worker_connections
+      worker_processes: "{{ .Values.nginx.workerProcesses }}"
+      enable_cpu_affinity: {{ default "true" .Values.nginx.enableCPUAffinity }}
+      worker_rlimit_nofile: {{ default 20480 .Values.nginx.workerRlimitNofile }}     # the number of files a worker process can open, should be larger than worker_connections
       event:
-        worker_connections: 10620
+        worker_connections: {{ default 10620 .Values.nginx.workerConnections  }}
       http:
         enable_access_log: {{ .Values.logs.enableAccessLog }}
         {{- if .Values.logs.enableAccessLog }}
         access_log: "{{ .Values.logs.accessLog }}"
-        access_log_format: "{{ .Values.logs.accessLogFormat }}"
+        access_log_format: '{{ .Values.logs.accessLogFormat }}'
         access_log_format_escape: {{ .Values.logs.accessLogFormatEscape }}
         {{- end }}
         keepalive_timeout: 60s         # timeout during which a keep-alive client connection will stay open on the server side.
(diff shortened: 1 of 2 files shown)


charts/apisix/templates/configmap.yaml
```diff
@@ -165,14 +165,16 @@ data:
     nginx_config:                     # config for render the template to genarate nginx.conf
       error_log: "{{ .Values.logs.errorLog }}"
       error_log_level: "{{ .Values.logs.errorLogLevel }}"         # warn,error
-      worker_rlimit_nofile: 20480     # the number of files a worker process can open, should be larger than worker_connections
+      worker_processes: "{{ .Values.nginx.workerProcesses }}"
```
fengxsong, line 168: oh my! duplicate to https://github.com/apache/apisix-helm-chart/pull/268

charts/apisix/values.yaml
```diff
@@ -186,6 +183,11 @@ admin:
     ipList:
       - 127.0.0.1/24
 
+nginx:
+  workerRlimitNofile: "1047552"
+  workerConnections: "16384"
```
tokers, line 188: These two fields should be integer?

fengxsong: nope, if those fields are number type, they will be converted into string type while helm command processing, for example, number 1047552 will be `1.047552e+06`, then cause apisix gateway fail to started.

tokers: Gotcha. I think this is due to the handling of number in helm.

charts/apisix/templates/configmap.yaml
```diff
@@ -165,14 +165,16 @@ data:
...
+      enable_cpu_affinity: {{ default "true" .Values.nginx.enableCPUAffinity }}
+      worker_rlimit_nofile: {{ default 20480 .Values.nginx.workerRlimitNofile }}     # the number of files a worker process can open, should be larger than worker_connections
       event:
-        worker_connections: 10620
+        worker_connections: {{ default 10620 .Values.nginx.workerConnections  }}
```
tokers, line 172: The `10620` should also add the double quote, to keep the consistence.

fengxsong: yes, make sense :)

charts/apisix/templates/configmap.yaml
```diff
@@ -165,14 +165,16 @@ data:
...
+      worker_rlimit_nofile: {{ default "20480" .Values.nginx.workerRlimitNofile }}     # the number of files a worker process can open, should be larger than worker_connections
       event:
-        worker_connections: 10620
+        worker_connections: {{ default "10620" .Values.nginx.workerConnections  }}
       http:
         enable_access_log: {{ .Values.logs.enableAccessLog }}
         {{- if .Values.logs.enableAccessLog }}
         access_log: "{{ .Values.logs.accessLog }}"
-        access_log_format: "{{ .Values.logs.accessLogFormat }}"
+        access_log_format: '{{ .Values.logs.accessLogFormat }}'
```
spacewander, line 177: Why change the quote?

fengxsong: single quote allows us not to escape value of `logs.accessLogFormat`, just `'{"remote_addr":"$remote_addr",...}'` not `'{\"remote_addr\":\"$remote_addr\",...}'`

tokers (approved):

charts/apisix/values.yaml
tao12345666333, line 188: we should keep the default value, right?

fengxsong: resolved

tao12345666333 (approved):
LGTM

Thanks

charts/apisix/templates/configmap.yaml
tao12345666333: @fengxsong After discussions with @spacewander , **we consider the modification here to be a breaking change.**

I will revert the changes here to make it consistent with the APISIX configuration file format.  https://github.com/apache/apisix/blob/7569eb99326f72f69a6e275126decce2b7e486d1/conf/config-default.yaml#L187

PR #316, opened 2022-07-26
feat: Allow `fullnameOverride`s to work correctly

NMichas:
Purpose: Allow `fullnameOverride`s to work correctly to fully automate Helm deployments.
Note: This is an updated PR, a lot less intrusive than the one suggested in https://github.com/apache/apisix-helm-chart/pull/292 (which I have now closed).

- In APISIX admin:
If the etcd subchart is configured to use a `fullnameOverride`, APISIX admin fails to start. This is due to the fact that etcd `fullnameOverride` is not taken into account in APISIX admin's init container nor configmap. This PR is backwards compatible in the sense that if users do not use a `fullnameOverride` for etcd, the chart behaves identically as before.

- In APISIX Ingress Controller:
`apisix.serviceName` and `apisix.serviceNamespace` must be explicitly set, thus not allowing users to fully automate a Helm install when using a custom `fullnameOverride` for APISIX admin which might have a different logic than the one currently used to construct the APISIX admin URL. This PR introduces a new value `apisix.serviceFullname`. When this value is set, it is overriding the logic of constructing the APISIX admin URL from `apisix.serviceName` and `apisix.serviceNamespace`. This PR is backwards compatible in the sense that if users do not use the newly introduced `apisix.serviceFullname` value, the chart behaves identically as before.

### DIFF
--- a/charts/apisix-ingress-controller/values.yaml
+++ b/charts/apisix-ingress-controller/values.yaml
@@ -123,7 +123,7 @@ config:
   # APISIX related configurations.
   apisix:
     # Enabling this value, overrides serviceName and serviceNamespace.
-    # serviceFullname: "apisix-admin"
+    # serviceFullname: "apisix-admin.apisix.svc.local"
     serviceName: apisix-admin
     serviceNamespace: ingress-apisix
     servicePort: 9180
--- a/charts/apisix-ingress-controller/values.yaml
+++ b/charts/apisix-ingress-controller/values.yaml
@@ -122,6 +122,8 @@ config:
     enableGatewayAPI: false
   # APISIX related configurations.
   apisix:
+    # Enabling this value, overrides serviceName and serviceNamespace.
+    # serviceFullname: "apisix-admin"
     serviceName: apisix-admin
     serviceNamespace: ingress-apisix
     servicePort: 9180
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .idea/
 .DS_Store
 test_**.yaml
+.history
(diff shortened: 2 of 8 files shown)


tao12345666333:
Great! 

let me take a look.

lingsamuel (approved):

charts/apisix-ingress-controller/values.yaml
```diff
@@ -122,6 +122,8 @@ config:
     enableGatewayAPI: false
   # APISIX related configurations.
   apisix:
+    # Enabling this value, overrides serviceName and serviceNamespace.
+    # serviceFullname: "apisix-admin"
```
tao12345666333, line 126: I think we should can set up an example here. e.g. `apisix-admin.apisix.svc.local`

tao12345666333 (commented):
LGTM

Leave a suggestion, and please update the documentation, thank you

tokers (approved):

charts/apisix-ingress-controller/values.yaml
NMichas: Done.

tao12345666333 (approved):

tao12345666333:
@NMichas Do you think we need to release a new version of the Helm chart?

NMichas:
> @NMichas Do you think we need to release a new version of the Helm chart?

Sorry for the late reply, I was on holiday. Yes, a new release is required.

tao12345666333:
@NMichas Don't worry.
If you have time, you can submit a PR that just needs to update its Helm chart version to release a new version. Thanks

NMichas:
@tao12345666333 sure, which versions should the update target? Current versions are:

APISIX:
- version: 0.11.1-rc1
- appVersion: 2.15.0

APISIX Ingress Controller:
- version: 0.10.0-rc1
- appVersion: 1.5.0-rc1

tao12345666333:
Sorry for delay. This update has been introduced to the latest version, no further action is required

PR #323, opened 2022-08-11
fix: Move ingressClassName to spec

jaysonsantos:
The property `ingressClassName` should be rendered inside `spec` and not `metadata`


tokers (approved):

tao12345666333 (approved):
LGTM

PR #327, opened 2022-08-17
feat: Allow user to toggle enableServerTokens from values

jaysonsantos:


douglasqsantosaudibene (approved):

charts/apisix/values.yaml
```diff
@@ -188,6 +188,8 @@ nginx:
   workerConnections: "10620"
   workerProcesses: auto
   enableCPUAffinity: true
+  # Whether apisix shows the full version on response header Server
+  enableServerTokens: true
```
tokers, line 192: It should not be in the nginx section.

jaysonsantos: hi there, could you point where it should be then?
`apisix` seemed too generic, `gateway` only configures the service

tokers: @jaysonsantos Just in APISIX.

PR #328, opened 2022-08-22
chore: support configuring SSL protocols

tokers:
Signed-off-by: Chao Zhang <tokers@apache.org>

### DIFF
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -142,6 +142,7 @@ gateway:
     certCAFilename: ""
     http2:
       enabled: true
+    sslProtocols: "TLSv1.2 TLSv1.3"
   # L4 proxy (TCP/UDP)
   stream:
     enabled: false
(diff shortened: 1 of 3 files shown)


charts/apisix/values.yaml
```diff
@@ -142,6 +142,7 @@ gateway:
     certCAFilename: ""
     http2:
       enabled: true
+    sslProtocols: "TLSv1.2 TLSv1.3"
```
bzp2010, line 145: It is different from the previous value, do we need to consider compatibility? 🤔

bzp2010 (approved):

charts/apisix/values.yaml
tao12345666333: I think we need.

tokers: Actually, I think we should drop the support of `TLSv1` and `TLSv1.1`, since both of them are unsafe and deprecated.

tao12345666333: > Actually, I think we should drop the support of `TLSv1` and `TLSv1.1`, since both of them are unsafe and deprecated.

I think it needs to be droped, but the behavior in the helm chart should be consistent with APISIX. Are these protocols droped in APISIX?

tokers: OK, then we can drop it after APISIX changes its default value.

tokers: See https://github.com/apache/apisix/blob/master/conf/config-default.yaml#L143.

tao12345666333: got it

tao12345666333 (approved):

PR #333, opened 2022-09-01
fix: incluster name resolve issue

fengxsong:
Signed-off-by: fengxsong <fengxsong@outlook.com>

When nginx listen with hostnetwork, change default `dnsPolicy` to `ClusterFirstWithHostNet`


charts/apisix/templates/_pod.tpl
```diff
@@ -115,6 +115,10 @@ spec:
       {{- end }}
       resources:
       {{- toYaml .Values.apisix.resources | nindent 8 }}
+  {{- if .Values.apisix.hostNetwork }}
```
tokers, line 118: Also add this field to readme.

fengxsong: done, maybe we can use tools like https://github.com/norwoodj/helm-docs to generate README later?

charts/apisix/templates/_helpers.tpl
```diff
@@ -96,17 +96,3 @@ prometheus:
 {{- define "apisix.pluginAttrs" -}}
 {{- merge .Values.pluginAttrs (include "apisix.basePluginAttrs" . | fromYaml) | toYaml -}}
 {{- end -}}
-
-{{- define "apisix.podAntiAffinity" -}}
```
tao12345666333, line 100: Why delete this one?

charts/apisix/templates/_pod.tpl
tao12345666333: Yes, I agree! Helm-docs is a good tool

charts/apisix/templates/_helpers.tpl
fengxsong: not need for this template function. as kube-scheduler will take care of this.

tokers (approved):

charts/apisix/templates/_helpers.tpl
tao12345666333: yes

tao12345666333 (approved):

PR #335, opened 2022-09-02
feat: create rbac resources when kubernetes service discovery enabled

nic-6443:
Signed-off-by: Nic <qianyong@api7.ai>

### DIFF
--- a/charts/apisix/templates/service-account.yaml
+++ b/charts/apisix/templates/service-account.yaml
@@ -20,5 +20,4 @@ kind: ServiceAccount
 metadata:
   name: {{ include "apisix.serviceAccountName" . }}
   namespace: {{ .Release.Namespace }}
-automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
 {{- end }}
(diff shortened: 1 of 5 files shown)


tokers (commented):
Also add the config reference in README.

charts/apisix/templates/service-account.yaml
```diff
@@ -0,0 +1,23 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+{{- if .Values.serviceAccount.create }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "apisix.serviceAccountName" . }}
```
tokers, line 21: Where did we define `apisix.serviceAccountName`?

nic-6443: https://github.com/apache/apisix-helm-chart/blob/2ebd372eff42ccf3e874cc016f00851e18b1ca79/charts/apisix/templates/_helpers.tpl#L54-L63

PR #345, opened 2022-09-13
Customizable ingressClassName for gateway

lpiob:
The helm chart supports exposing Apisix Gateway Service through an Ingress.

Unfortunately it does not support choosing `ingressClassName`, so it would work only if there's a default available.

This PR allows setting up a custom Ingress `className` within `values.yaml`, as for example:
```
gateway:
  ingress:
    enabled: true
    className: alb-private
    hosts:
      - host: api.dev.project.int.corp
        paths:
          - /
```

Variable name was choosen to be consisting with className used in apisix-dashboard. The change was tested and is backwards compatible.

### DIFF
--- a/charts/apisix/README.md
+++ b/charts/apisix/README.md
@@ -96,7 +96,7 @@ Apache APISIX service parameters, this determines how users can access itself.
 | `gateway.stream`                | Apache APISIX service settings for stream                                                                                                                                           |            |
 | `gateway.ingress`               | Using ingress access Apache APISIX service                                                                                                                                          |            |
 | `gateway.ingress.annotations`   | Ingress annotations                                                                                                                                                                 | `[]`       |
-| `gateway.ingress.className`     | `ingressClassName` replaces `annotations kubernetes.io/ingress.class`, required kubernetes 1.18>=                                                                                   |            |
+| `gateway.ingress.className`     | `ingressClassName` replaces `annotations kubernetes.io/ingress.class`, required Kubernetes `1.18>=`                                                                                   |            |
 
 ### admin parameters
 
--- a/charts/apisix/README.md
+++ b/charts/apisix/README.md
@@ -95,6 +95,8 @@ Apache APISIX service parameters, this determines how users can access itself.
 | `gateway.tls.sslProtocols`    |   TLS protocols allowed to use.  | `"TLSv1.2 TLSv1.3"`       |
 | `gateway.stream`                | Apache APISIX service settings for stream                                                                                                                                           |            |
 | `gateway.ingress`               | Using ingress access Apache APISIX service                                                                                                                                          |            |
+| `gateway.ingress.annotations`   | Ingress annotations                                                                                                                                                                 | `[]`       |
+| `gateway.ingress.className`     | `ingressClassName` replaces `annotations kubernetes.io/ingress.class`, required kubernetes 1.18>=                                                                                   |            |
 
 ### admin parameters
 
--- a/charts/apisix/README.md
+++ b/charts/apisix/README.md
@@ -96,7 +96,7 @@ Apache APISIX service parameters, this determines how users can access itself.
 | `gateway.stream`                | Apache APISIX service settings for stream                                                                                                                                           |            |
 | `gateway.ingress`               | Using ingress access Apache APISIX service                                                                                                                                          |            |
 | `gateway.ingress.annotations`   | Ingress annotations                                                                                                                                                                 | `[]`       |
-| `gateway.ingress.className`     | `ingressClassName` replaces `annotations kubernetes.io/ingress.class`, required Kubernetes `1.18>=`                                                                                   |            |
+| `gateway.ingress.className`     | `ingressClassName` replaces `annotations kubernetes.io/ingress.class`, required Kubernetes `>=1.18`                                                                                   |            |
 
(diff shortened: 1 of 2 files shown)


charts/apisix/templates/ingress.yaml
```diff
@@ -34,6 +34,9 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
+  {{- if .Values.gateway.ingress.className }}
+  ingressClassName: {{ .Values.gateway.ingress.className }}
```
tokers, line 38: This field was introduced since k8s 1.18, but the ingress resource for apisix is also compatible with older ingress version (extensions). So only add this field only if the k8s version is new enough.

We also need to consider removing the support for older ingress versions.

lpiob: Condition modified to the suggestion.

charts/apisix/README.md
```diff
@@ -95,6 +95,8 @@ Apache APISIX service parameters, this determines how users can access itself.
 | `gateway.tls.sslProtocols`    |   TLS protocols allowed to use.  | `"TLSv1.2 TLSv1.3"`       |
 | `gateway.stream`                | Apache APISIX service settings for stream                                                                                                                                           |            |
 | `gateway.ingress`               | Using ingress access Apache APISIX service                                                                                                                                          |            |
+| `gateway.ingress.annotations`   | Ingress annotations                                                                                                                                                                 | `[]`       |
+| `gateway.ingress.className`     | `ingressClassName` replaces `annotations kubernetes.io/ingress.class`, required kubernetes 1.18>=                                                                                   |            |
```
tokers, line 99:
```suggestion
| `gateway.ingress.className`     | `ingressClassName` replaces `annotations kubernetes.io/ingress.class`, required Kubernetes `1.18>=`                                                                                   |            |
```

charts/apisix/templates/ingress.yaml
```diff
@@ -34,7 +34,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  {{- if .Values.gateway.ingress.className }}
+  {{- if and .Values.gateway.ingress.className (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion) }}
```
tokers, line 37: Then for old version, do we need to add the class annotation?

lpiob: Implemented now in <https://github.com/apache/apisix-helm-chart/pull/345/files#diff-3a22eeb01325380f68706901c6733b574958c40d0b7329385729f85ba4a74c80R20>, in the same way that [current helm](https://github.com/helm/helm/blob/bed23120b048977d67c929cec915e1d6e0887c5d/pkg/chartutil/create.go#L218-L242) templates adds those annotations.

The change was tested by simulating older kube version:
```
l-lbiegaj apisix [gateway-]≠1?5 -> helm template . --set 'gateway.ingress.enabled=true' --set 'gateway.ingress.className=TESTME' --kube-version 1.17  | grep -A15 Ingress
kind: Ingress
metadata:
  name: RELEASE-NAME-apisix
  labels:
    helm.sh/chart: apisix-0.11.1-localtests
    app.kubernetes.io/name: apisix
    app.kubernetes.io/instance: RELEASE-NAME
    app.kubernetes.io/version: "2.15.0"
    app.kubernetes.io/managed-by: Helm
  annotations:
    kubernetes.io/ingress.class: TESTME
spec:
  rules:
...

l-lbiegaj apisix [gateway-]≠1?5 -> helm template . --set 'gateway.ingress.enabled=true' --set 'gateway.ingress.className=TESTME' --kube-version 1.23  | grep -A15 Ingress
kind: Ingress
metadata:
  name: RELEASE-NAME-apisix
  labels:
    helm.sh/chart: apisix-0.11.1-localtests
    app.kubernetes.io/name: apisix
    app.kubernetes.io/instance: RELEASE-NAME
    app.kubernetes.io/version: "2.15.0"
    app.kubernetes.io/managed-by: Helm
spec:
  ingressClassName: TESTME
  rules:
...
```

tokers (approved):

TomekKrasuski (approved):

tao12345666333 (approved):

charts/apisix/README.md
```diff
@@ -95,6 +95,8 @@ Apache APISIX service parameters, this determines how users can access itself.
 | `gateway.tls.sslProtocols`    |   TLS protocols allowed to use.  | `"TLSv1.2 TLSv1.3"`       |
 | `gateway.stream`                | Apache APISIX service settings for stream                                                                                                                                           |            |
 | `gateway.ingress`               | Using ingress access Apache APISIX service                                                                                                                                          |            |
+| `gateway.ingress.annotations`   | Ingress annotations                                                                                                                                                                 | `[]`       |
+| `gateway.ingress.className`     | `ingressClassName` replaces `annotations kubernetes.io/ingress.class`, required Kubernetes `1.18>=`                                                                                   |            |
```
tao12345666333, line 99:
```suggestion
| `gateway.ingress.className`     | `ingressClassName` replaces `annotations kubernetes.io/ingress.class`, required Kubernetes `>=1.18`                                                                                   |            |
```

PR #347, opened 2022-09-15
change: sync config of APISIX V3 version

tzssangglass:
In APISIX V3 version, there are some breaking changes in `config-default.yaml`, and this PR is changed synchronously to this repo.

The following are the PRs I have filtered in relation to breaking changes: 
https://github.com/apache/apisix/pull/7716
https://github.com/apache/apisix/pull/7717
https://github.com/apache/apisix/pull/7806
https://github.com/apache/apisix/pull/7860
https://github.com/apache/apisix/pull/7823
https://github.com/apache/apisix/pull/7901 (in review)

I am not familiar with the helm chart, please let me know if there is anything else that needs to be changed.

### DIFF
--- a/charts/apisix/templates/configmap.yaml
+++ b/charts/apisix/templates/configmap.yaml
@@ -62,12 +62,10 @@ data:
       enable_dev_mode: false                       # Sets nginx worker_processes to 1 if set to true
       enable_reuseport: true                       # Enable nginx SO_REUSEPORT switch if set to true.
       enable_ipv6: {{ .Values.apisix.enableIPv6 }} # Enable nginx IPv6 resolver
-      config_center: etcd                          # etcd: use etcd to store the config value
-                                                   # yaml: fetch the config value from local yaml file `/your_path/conf/apisix.yaml`
 
       #proxy_protocol:                 # Proxy Protocol configuration
-      #  listen_http_port: 9181        # The port with proxy protocol for http, it differs from node_listen and port_admin.
-                                      # This port can only receive http request with proxy protocol, but node_listen & port_admin
+      #  listen_http_port: 9181        # The port with proxy protocol for http, it differs from node_listen and admin_listen.
+                                      # This port can only receive http request with proxy protocol, but node_listen & admin_listen
                                       # can only receive http request. If you enable proxy protocol, you must use this port to
                                       # receive http request with proxy protocol
       #  listen_https_port: 9182       # The port with proxy protocol for https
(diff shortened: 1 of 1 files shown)


nic-6443 (approved):

tokers:
I have a concern that once we synced the content of `config.yaml` to v3, how can we improve the `config.yaml` for APISIX `v2.x`? IMHO the content of `config.yaml` in the ConfigMap may still lose some content in comparison to the one in APISIX repo. We don't have a release branch to allow us to iterate released versions.

Also, v3 is brand new and it's not uncommon that in a long while a lot of users will still use APISIX v2. It's worth paying more attention on this, not just syncing the differences.

moonming:
> 

can we add `version` when installing apisix?

tzssangglass:
There should be two versions, corresponding to V2 and V3 of APISIX?

tokers:
> There should be two versions, corresponding to V2 and V3 of APISIX?

Yes.

tao12345666333:
> I have a concern that once we synced the content of `config.yaml` to v3, how can we improve the `config.yaml` for APISIX `v2.x`? IMHO the content of `config.yaml` in the ConfigMap may still lose some content in comparison to the one in APISIX repo. We don't have a release branch to allow us to iterate released versions.
> 
> Also, v3 is brand new and it's not uncommon that in a long while a lot of users will still use APISIX v2. It's worth paying more attention to this, not just syncing the differences.
> 
> cc @tao12345666333 @moonming

Maybe we can maintain a new branch dedicated to releasing v3 related Helm charts.
But this causes users to install v3 silently when installing the latest version.

I suggest that instead of maintaining a specific branch, we can directly add a configuration item to complete its compatibility.

tokers:
> > I have a concern that once we synced the content of `config.yaml` to v3, how can we improve the `config.yaml` for APISIX `v2.x`? IMHO the content of `config.yaml` in the ConfigMap may still lose some content in comparison to the one in APISIX repo. We don't have a release branch to allow us to iterate released versions.
> 
> > 
> 
> > Also, v3 is brand new and it's not uncommon that in a long while a lot of users will still use APISIX v2. It's worth paying more attention to this, not just syncing the differences.
> 
> > 
> 
> > cc @tao12345666333 @moonming
> 
> 
> 
> Maybe we can maintain a new branch dedicated to releasing v3 related Helm charts.
> 
> But this causes users to install v3 silently when installing the latest version.
> 
> 
> 
> I suggest that instead of maintaining a specific branch, we can directly add a configuration item to complete its compatibility.

You mean, translating the configuration according to the apisix version?

tao12345666333:
It can be like this:

```
# Add option for APISIX major version.

if apisix_major_version == 'v3'
    # using v3 configuration template    
else
     # using v2 configuration template
```

Or we can add some translate logic in Chart.
I think this way the user experience will be better, but this may also lead to some confusion for users.

Typically, users prefer the configuration item to be the same as the actual configuration used.

tzssangglass:
any news?

tao12345666333:
I tend to use the first way.
If we want to continue, we should first reach a consensus

tzssangglass:
what are your options? cc @tokers @nic-6443 @moonming @gxthrj

tokers:
The `apisix_major_version` can be detected from the APISXI image? We don't need to let our users do the choice.

tao12345666333:
> The `apisix_major_version` can be detected from the APISXI image? We don't need to let our users do the choice.

In this case, it means that APISIX is required to complete the compatibility of related configurations?

tzssangglass:
> The `apisix_major_version` can be detected from the APISXI image? We don't need to let our users do the choice.

here: https://github.com/apache/apisix/blob/master/conf/config-default.yaml#L588

tzssangglass:
> In this case, it means that APISIX is required to complete the compatibility of related configurations?

In fact APISIX is hardly compatible and some modifications are inherently breaking change.

tokers:
> > The `apisix_major_version` can be detected from the APISXI image? We don't need to let our users do the choice.
> 
> here: https://github.com/apache/apisix/blob/master/conf/config-default.yaml#L588

I mean, from the APISIX image tag, we can know the version. Then we can decide how to render the config.

tzssangglass:
> I mean, from the APISIX image tag, we can know the version.

Use an image with a 3.0 version number after the 3.0 release?
ref: https://hub.docker.com/r/apache/apisix/tags

tokers:
> > I mean, from the APISIX image tag, we can know the version.
> 
> Use an image with a 3.0 version number after the 3.0 release? ref: https://hub.docker.com/r/apache/apisix/tags

Yep.

tao12345666333:
you can use `semverCompare` for this one https://helm.sh/docs/chart_template_guide/function_list/#semvercompare

and `if/else` flow control https://helm.sh/docs/chart_template_guide/control_structures/#ifelse

PR #360, opened 2022-09-30
Customizable http router name

nicolaracco:
It would be helpful to allow the customization of the http router without needing to override the entire config


charts/apisix/values.yaml
```diff
@@ -56,6 +56,8 @@ apisix:
         - key: ""
           path: ""
 
+  httpRouter: radixtree_uri
```
tao12345666333, line 59: I think it would be better to add comments to here. 

list all options

tao12345666333 (commented):
Thanks 

Please update chart's README as well

nicolaracco:
Hello @tao12345666333. I've updated both the values.yml and the README

tao12345666333:
I mean this one https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/README.md

nicolaracco:
Sorry :) I've updated the right file this time

tao12345666333 (approved):
LGTM

nic-6443 (approved):

PR #372, opened 2022-10-14
fix: use https scheme when etcd.auth.tls.enabled

hadican:
Eventhough the `etcd.auth.tls.enabled`, `apisix` tries to connect with `http`.

This PR will solve it.

### DIFF
--- a/charts/apisix/templates/configmap.yaml
+++ b/charts/apisix/templates/configmap.yaml
@@ -286,9 +286,9 @@ data:
           {{- else }}
           host:                                 # it's possible to define multiple etcd hosts addresses of the same etcd cluster.
             {{- if .Values.etcd.fullnameOverride }}
-            - "http://{{ .Values.etcd.fullnameOverride }}:{{ .Values.etcd.service.port }}"
+            - "{{ include "apisix.etcd.auth.scheme" . }}://{{ .Values.etcd.fullnameOverride }}:{{ .Values.etcd.service.port }}"
             {{- else }}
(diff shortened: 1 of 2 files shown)


charts/apisix/templates/configmap.yaml
```diff
@@ -276,12 +276,21 @@ data:
               role: viewer
         etcd:
         {{- if .Values.etcd.enabled }}
+          {{- if .Values.etcd.auth.tls.enabled }}
+          host:                                 # it's possible to define multiple etcd hosts addresses of the same etcd cluster.
+            {{- if .Values.etcd.fullnameOverride }}
+            - "https://{{ .Values.etcd.fullnameOverride }}:{{ .Values.etcd.service.port }}"
```
tokers, line 282: Is it possible to define a `scheme` variable here (see https://helm.sh/docs/chart_template_guide/variables/), so that we can reduce the template complexity?

tao12345666333: yes

hadican: How about a helper like:
```
{{/*
Scheme to use while connecting etcd
*/}}
{{- define "apisix.etcd.auth.scheme" -}}
{{- if .Values.etcd.auth.tls.enabled }}
{{- "https" }}
{{- else }}
{{- "http" }}
{{- end }}
{{- end }}
```

tokers: Cool.

tokers (approved):

tao12345666333 (approved):

PR #366, opened 2022-10-10
docs: How to install Apache APISIX running in standalone mode?

oil-oil:
Added a new FAQ on _How to install Apache APISIX running in standalone mode._
Screenshot of the Yaml file display: 
temp

PR #375, opened 2022-10-18
add extraInitContainers

ohayak:
I added an optional parameter to configure extra initContainers

the purpose of this option is to give user an option to do more customization, in my case it was to patch config files in order to support Stand-alone mode. Stand-alone mode won't be easily deployable before v3.0.0 but for people using v2 it's tricky to make it work without extra customization.


tao12345666333:
Can you add more detailed background to the description? Why do we need this? What problem can it solve? thanks

ohayak:
> Can you add more detailed background to the description? Why do we need this? What problem can it solve? thanks

description updated

tokers:
> I added an optional parameter to configure extra initContainers
> 
> 
> 
> the purpose of this option is to give user an option to do more customization, in my case it was to patch config files in order to support Stand-alone mode. Stand-alone mode won't be easily deployable before v3.0.0 but for people using v2 it's tricky to make it work without extra customization.

Interesting, I think the way is worthy to be written in the docs. Would you also update the docs to introduce the way you patch the config to support the standalone mode.

tao12345666333:
> in my case it was to patch config files in order to support Stand-alone mode.

Could you update the documentation to add a specific description about it?

There is currently a similar document describing stand-alone mode
https://github.com/apache/apisix-helm-chart/pull/366

g-javi:
This is also helpful when changing ownership of mounted volumes (e.g. hostpath) when the `file-logger` plugin is used.

g-javi (approved):
LGTM!

g-javi:
@tao12345666333 @tokers any update on this?

lingsamuel (approved):

tokers (approved):

tao12345666333 (approved):

PR #458, opened 2023-02-03
fix: include grpc-web in default plugins

fracek:
I noticed that grpc-web is not included in the default plugin list.


tao12345666333 (approved):

fracek:
I updated the PR to fix ci.

tokers (approved):

tao12345666333 (approved):

PR #462, opened 2023-02-07
docs:Fix FAQ standalone mode - use debian instead of obsolete alpine,…

sbrouet:
Issue 454 doc: fix standalone mode FAQ
Fixes following items of issue:
* Issue 1: sample Kubernetes files out of date
* Issue 3: add recommendation for Linux distribution to be used
Thanks

### DIFF
--- a/docs/en/latest/FAQ.md
+++ b/docs/en/latest/FAQ.md
@@ -42,16 +42,17 @@ helm install apisix apisix/apisix \
 Helm chart does not provide a direct way to deploy Apache APISIX running in standalone mode. You can install it in the following manner.
 
 Create a `deploy.yaml` with the following content.
+It is recommended using same Linux distribution as is used by the Helm Chart, which currently is debian.
 
 <Tabs
-  groupId="version"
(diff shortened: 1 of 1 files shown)


tao12345666333 (commented):
Thanks for your contribution.

Updating to the latest version is what it should be.

Did you try the installation with the latest version and it works?

docs/en/latest/FAQ.md
```diff
@@ -167,7 +168,7 @@ spec:
 
 </TabItem>
 
-<TabItem value="2.15">
+<TabItem value="3.1.0-debian">
```
AlinsRan, line 171: The configmap configuration here is incorrect. The 3.x.x version uses the 2.x.x version. I think you should install and verify it.

AlinsRan (changes requested):

AlinsRan:
Due to the different configurations of 2.x.x and 3.x.x, I suggest that `2.15.x-alpine` and `3.x.x-centos` and `3.x.x-debian` be reserved at the same time.

Issue #463, opened 2023-02-07
hpa version will deprecate


jiangfucheng:
Hi, hpa version of apisix, apisix-dashboard and ingress-controller is v2bate1, this version is deprecated in new k8s version. So that i want upgrade hpa version from v2bate1 to v2. Please assign to me if think ok.

tokers:
> Hi, hpa version of apisix, apisix-dashboard and ingress-controller is v2bate1, this version is deprecated in new k8s version. So that i want upgrade hpa version from v2bate1 to v2. Please assign to me if think ok.

We consider being compatible with the latest few K8s release (maybe the latest four or five). As long as the `v2` version si available among all of these releases, I think upgrade is OK.

tao12345666333:
Assigned.

I think you can add a configuration item,
Allow users to configure the API version that can be used.

PR #468, opened 2023-02-12
feat: support v2 version for hpa

jiangfucheng:
Fix https://github.com/apache/apisix-helm-chart/issues/463


jiangfucheng:
@tao12345666333  @tokers  PTAL

tokers (commented):
Do we need to add a auto selection feature for users. That is, if the target Kubernetes cluster is appropriate, let's use v2, or let's use `v2beta1`.

tao12345666333 (changes requested):
There is no need to introduce a new template file.
You can directly add conditional judgments to the original template file.

For `autoscaling/v2` and `autoscaling/v2beta1` the changes are not that big enough to warrant a new template file.

jiangfucheng:
> Do we need to add a auto selection feature for users. That is, if the target Kubernetes cluster is appropriate, let's use v2, or let's use `v2beta1`.

Is it feasible to use `.Capabilities.APIVersions.Has "autoscaling/v2"` to check hpa version？

jiangfucheng:
Sorry for delay, I just add a version variable to select the version of HPA, since i find the `.Capabilities.APIVersions.Has` maybe can't check whether k8s cluster supports this version, it just use the version of Helm as check condition. Please review again, @tao12345666333 @tokers .

charts/apisix-ingress-controller/README.md
```diff
@@ -109,6 +109,7 @@ The same for container level, you need to set:
 | autoscaling.maxReplicas | int | `100` |  |
 | autoscaling.minReplicas | int | `1` |  |
 | autoscaling.targetCPUUtilizationPercentage | int | `80` |  |
+| autoscaling.version | string | `"v2"` | HPA version, default v2 |
```
tokers, line 112: Also show the candidate versions here.

jiangfucheng: Done.

tao12345666333 (approved):

gkalwig:
Hi, when you plan release that change? We are unable to update our environment without it, and we really need it.

Gallardot (approved):

Issue #481, opened 2023-02-24
The deployment of control-plane should support closing the gateway option


sggene:
At present, apisix-control-plane-gateway cannot provide services and has no significance
```
values-cp.yaml 
gateway:
  `enable: false`
  # -- Apache APISIX service type for user access itself
  type: NodePort
...
```

tokers:
@jinjianming PR's welcome!

jinjianming:
> @jinjianming欢迎 PR！

https://github.com/apache/apisix-helm-chart/pull/482

PR #482, opened 2023-02-24
feat: Add `gateway.enabled` to adjust in decoupled mode

sggene:


jinjianming:
https://github.com/apache/apisix-helm-chart/issues/481

charts/apisix/values.yaml
```diff
@@ -233,11 +234,13 @@ gateway:
     enabled: true
     servicePort: 80
     containerPort: 9080
+    nodePort: ""
```
tokers, line 237: When using the empty string as the default value of `nodePort`, it might mislead users to configure a string for `nodePort`. `nodePort` should be an integer, let's just comment the `nodePort` when we don't need it.

jinjianming: Changes have been made, comments have been added and resubmitted

PR #490, opened 2023-02-28
bugfix: 0.0.0.0/0 is always set for allow_admin

igor-borisoglebski:
0.0.0.0/0 is always set for allow_admin. If this is intentional, I suggest this PR is dismissed, but that the documentation is updated registering that fact as it is misleading in its current form

### DIFF
--- a/charts/apisix/templates/configmap.yaml
+++ b/charts/apisix/templates/configmap.yaml
@@ -281,9 +281,6 @@ data:
         {{- else }}
           - 0.0.0.0/0
         {{- end}}
-        {{- if or (index .Values "ingress-controller" "enabled") .Values.dashboard.enabled  }}
-          - 0.0.0.0/0
-        {{- end}}
         #   - "::/64"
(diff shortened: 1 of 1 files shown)


tokers:
@igor-borisoglebski This is intentional for convenience, or you have to configure the IP address of the dashboard or ingress controller on the APISIX, which means you need to deploy the ingress controller or dashboard before APISIX. While they before they run, they need to wait until APISIX is ready.

That's the chicken and egg's situation.

igor-borisoglebski:
I've tested that the ingress comes up even when 0.0.0.0/0 is not set. The reason it does is because ingress checks if the port is listening with netcat - which succeeds - even though the API access is not allowed.

tao12345666333:
This is by design. The ingress controller needs to access the Admin API.
This is convenient for most users.
But if you need a more secure configuration, you can modify it yourself, or add NetworkPolicy to control it.

I will close this one since we're not going to tweak it. Thanks!

PR #491, opened 2023-02-28
doc: Documenting the possibility to specify nodePort for HTTP and TLS

igor-borisoglebski:
Documenting the possibility to specify nodePort for HTTP and TLS. This is useful when deploying in NodePort mode


lingsamuel:
The doc is auto generated. Comments should be placed in values.yaml

PR #492, opened 2023-02-28
feat: adding IPv6 Single Stack support

igor-borisoglebski:
Currently the Helm chart doesn't works on IPv6 single stack as the apisix-ingress-controller will wait until apisix-admin is serving. However, currently it is not possible to set the apisix-admin listen address given 0.0.0.0 is hardcoded and in that way it will never work properly on IPv6 single stack deployments. 

This PR adds the possiblity to set the apisix-admin IP listen address to any value. It should be set to `[::]` when on IPv6 single stack deployments.


tokers (approved):

tao12345666333:
Please check the CI job error.

tao12345666333:
you should run `helm-docs`

igor-borisoglebski:
@tao12345666333 fixed. Many thanks

tao12345666333 (approved):

PR #510, opened 2023-03-14
feat: support set etcd persistence

onemore5s:
add custom etcd persistence support to `apisix/values.yaml`

### DIFF
--- a/charts/apisix/README.md
+++ b/charts/apisix/README.md
@@ -134,7 +134,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | dns.resolvers[5] | string | `"8.8.8.8"` |  |
 | dns.timeout | int | `5` |  |
 | dns.validity | int | `30` |  |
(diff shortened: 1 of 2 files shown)


Gallardot:
Why do we need to maintain default values for etcd charts here  ? If users need a custom etcd configuration, it is better to guide users through the documentation for etcd charts.  https://github.com/bitnami/charts/tree/main/bitnami/etcd

sweetpotatoman:
> Why do we need to maintain default values for etcd charts here ? If users need a custom etcd configuration, it is better to guide users through the documentation for etcd charts. https://github.com/bitnami/charts/tree/main/bitnami/etcd

It is in order to see the relevant configuration of etcd persistence in `apisix/values.yaml`. 
There is no need to go to https://github.com/bitnami/charts/tree/main/bitnami/etcd to check.

I think etcd persistence is more commonly used, so I want to add it, just like `etcd.host` | `etcd.replicaCount`.

PR #516, opened 2023-03-28
fix: disable default cpu affinity

bzp2010:
Turn off the `worker_cpu_affinity` default switch for APISIX in Kubernetes to avoid performance anomalies under high load.


tao12345666333:
Do you have some specific stress test data and scenarios to share?

Issue #536, opened 2023-04-13
This configuration may cause admin. allow_admin completely fails


flylan:
When .Values.ingress-controller.enabled or .Values.dashboard.enabled is true, an IP segment of 0.0.0.0/0 will be added by default. However, this 0.0.0.0/0 configuration will make all clients accessible, equivalent to admin.allow_admin configuration becomes invalid

![image](https://user-images.githubusercontent.com/12097229/231643597-dc7d6092-1b58-452d-a53c-7e22a7244bb3.png)

I think changing it to this would be okay

![image](https://user-images.githubusercontent.com/12097229/231644280-04d900a2-1729-49c6-bdc5-758339d883da.png)

After all, the internal IPs of K8S are all private IPs, and all you need to do is allow all IPs to access them

flylan:
I submitted an optimization fix https://github.com/apache/apisix-helm-chart/pull/537

PR #537, opened 2023-04-13
Optimize admin.allow_admin configuration generation logic

flylan:
1. modify admin.allow_admin configured IP list logic to reduce the generation of duplicate IP segment configurations

2. when ingress-controller or dashboard is enabled, all private IP segments are added by default instead of 0.0.0.0/0 to improve security

### DIFF
--- a/charts/apisix/templates/configmap.yaml
+++ b/charts/apisix/templates/configmap.yaml
@@ -287,16 +287,16 @@ data:
 
       admin:
         allow_admin:    # http://nginx.org/en/docs/http/ngx_http_access_module.html#allow
+        {{- $ipList := list "0.0.0.0/0" -}}
         {{- if .Values.admin.allow.ipList }}
-        {{- range $ips := .Values.admin.allow.ipList }}
+        {{- $ipList = .Values.admin.allow.ipList -}}
+        {{- end }}
+        {{- if and (or (index .Values "ingress-controller" "enabled") .Values.dashboard.enabled) (not (has "0.0.0.0/0" $ipList)) }}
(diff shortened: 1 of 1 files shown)


flylan:
relate：https://github.com/apache/apisix-helm-chart/issues/536

charts/apisix/templates/configmap.yaml
```diff
@@ -287,16 +287,16 @@ data:
...
-          - {{ $ips }}
+        {{- $ipList = .Values.admin.allow.ipList -}}
         {{- end }}
-        {{- else }}
-          - 0.0.0.0/0
-        {{- end}}
-        {{- if or (index .Values "ingress-controller" "enabled") .Values.dashboard.enabled  }}
-          - 0.0.0.0/0
+        {{- if and (or (index .Values "ingress-controller" "enabled") .Values.dashboard.enabled) (not (has "0.0.0.0/0" $ipList)) }}
+        {{- $ipList = concat $ipList ( list "127.0.0.1/32" "10.0.0.0/8" "172.16.0.0/12" "192.168.0.0/16" ) | uniq -}}
```
tao12345666333, line 295: If security is a concern, we use NetworkPolicy better, right？

Sn0rt: @flylan PTAL

tao12345666333: And in some companies that I know of, there are indeed cases where non-private network segments are used as private networks.

flylan: Indeed, some companies may directly use public IP as the K8S networking environment, such as when multiple subsidiaries are distributed in different places

PR #546, opened 2023-05-24
adding fallback_sni to apisix conf

apalapal:
Some "clients", like AWS Application Load balancer, when forwarding HTTPS requests to upstream servers ( like APISIX ) do not present SNI on the TLS handshake, resulting on APISIX not to be able to fetch the appropriate certificate to establish the connection.
This change is to be able to define and include the "fallback_sni" property on apisix configuration, so that a last resort SNI can be defined and a proper certificate can be retrieved.

### DIFF
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -258,6 +258,8 @@ gateway:
       enabled: true
     # -- TLS protocols allowed to use.
     sslProtocols: "TLSv1.2 TLSv1.3"
+    # -- Define sni to fallback if none is presented by client
+    fallbackSni: "" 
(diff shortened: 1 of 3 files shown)


charts/apisix/values.yaml
```diff
@@ -258,6 +258,8 @@ gateway:
       enabled: true
     # -- TLS protocols allowed to use.
     sslProtocols: "TLSv1.2 TLSv1.3"
+    # -- Define sni to fallback if none is presented by client
+    fallbackSni: ""
```
tao12345666333, line 262: Let's rename it to fallbackSNI

Gallardot (approved):

tao12345666333 (approved):
LGTM

Issue #545, opened 2023-05-23
feat: Customize the readiness probe of APISIX pod


joyyir:
The current version of APISIX helm chart cannot change the readiness probe of APISIX pod. ([Source Code](https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/templates/_pod.tpl#L108))
It would be nice to support customizing the readiness probe of APISIX pod. It is probably needed to change the readiness probe, especially when using external plugins. Currently, the 10 second delay is not sufficient to prepare external plugins.

Gallardot:
Agree. Welcome to submit a PR to make it possible!

lingsamuel:
Closing since the PR merged

PR #548, opened 2023-05-31
feat: Customize the readiness probe of APISIX pod (#545)

joyyir:
Please refer #545

### DIFF
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -157,6 +157,16 @@ apisix:
   #         key: KEY
   extraEnvVars: []
 
+  # -- Set the readinessProbe for Apache APISIX pods
+  readinessProbe:
+    failureThreshold: 6
+    initialDelaySeconds: 10
+    periodSeconds: 10
+    successThreshold: 1
+    tcpSocket:
+      port: &gatewayHttpContainerPort 9080
+    timeoutSeconds: 1
+
 nameOverride: ""
 fullnameOverride: ""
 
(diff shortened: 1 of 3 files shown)


charts/apisix/values.yaml
```diff
@@ -157,6 +157,16 @@ apisix:
...
   extraEnvVars: []
 
+  # -- Set the readinessProbe for Apache APISIX pods
+  readinessProbe:
+    failureThreshold: 6
+    initialDelaySeconds: 10
+    periodSeconds: 10
+    successThreshold: 1
+    tcpSocket:
+      port: &gatewayHttpContainerPort 9080
```
tao12345666333, line 167: This is incorrect.

charts/apisix/values.yaml
```diff
@@ -232,7 +242,7 @@ gateway:
   http:
     enabled: true
     servicePort: 80
-    containerPort: 9080
+    containerPort: *gatewayHttpContainerPort
```
tao12345666333, line 245: ditto

tao12345666333 (changes requested):

charts/apisix/values.yaml
joyyir: I used `YAML Anchors` to refer to the same values.
Wouldn't you recommend it?
https://helm.sh/docs/chart_template_guide/yaml_techniques/#yaml-anchors.

tao12345666333: Sorry, this does work. I may have made a mistake during my previous verification.

tao12345666333 (approved):
LGTM

lingsamuel (approved):

Gallardot (approved):

PR #549, opened 2023-06-01
fix: config map admin/viewer secret key

apalapal:
The implementation of admin/view keys using kubernetes secret on #530 doesn't seem to work.
What happens is that on the resulting config.yaml file, the admin key ( and viewer ) will look like:
      - name: "admin"
        key: "{{APISIX_ADMIN_KEY}}" 

instead of the ' key: ${{APISIX_ADMIN_KEY}} '   that is described on [https://apisix.apache.org/docs/apisix/admin-api/](https://apisix.apache.org/docs/apisix/admin-api/).
The result is that the admin key is literally the value string above and not the value from the env/k8s secret.
-
This PR is the fix on the config map to have the proper key value format to reference the env variable ( filled from the configured secretName secret).

This change was tested, where the admin and viewer keys were correctly set and looked like this on the final config.yaml:

...
      - name: "admin"
        key: ${{APISIX_ADMIN_KEY}}
...
      - name: "viewer"
        key: ${{APISIX_VIEWER_KEY}}

### DIFF
--- a/charts/apisix/templates/configmap.yaml
+++ b/charts/apisix/templates/configmap.yaml
@@ -315,7 +315,7 @@ data:
           # admin: can everything for configuration data
           - name: "admin"
             {{- if .Values.admin.credentials.secretName }}
-            key: "{{"{{"}}APISIX_ADMIN_KEY{{"}}"}}"
+            key: {{"${{"}}APISIX_ADMIN_KEY{{"}}"}}
             {{- else }}
             key: {{ .Values.admin.credentials.admin }}
             {{- end }}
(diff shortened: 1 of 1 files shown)


Gallardot (approved):

ikatlinsky:
nice, found the same problem and fixed it as a part of https://github.com/apache/apisix-helm-chart/pull/551

apalapal:
> nice, found the same problem and fixed it as a part of #551

Indeed :-) I wish I had found yours before spending some time troubleshooting it :-) but I guess we were doing it more or less at the same time.
I would actually cancel this one, because yours not only fixes the problem as also adds new functionality for overriding the keys, but as it's already submitted, I'll leave it to the reviews to comment...
Great work by the way...

Zebradil:
It'd be really nice to get this bugfix merged and released.

lingsamuel (approved):

tao12345666333 (approved):

Issue #552, opened 2023-06-02
Allow to override admin key for ingress controller using secrets


ikatlinsky:
**Motivation**

Currently, there is a way to provide ingress controller `config.apisix.adminKey` only via plain text and it is not synced with the corresponding admin role token in the gateway itself.

**Proposal**

Allow providing name of the secret and key of the value in secret which should be used as an admin token to access gateway admin API:

```yaml
config:
  apisix:
      adminKey: ""
  
      # -- The APISIX Helm chart supports storing user credentials in a secret.
      # The secret needs to contain a single key for admin token with key adminKey by default.
      secretName: "gateway-secrets"
      # -- Name of the admin token key in the secret, overrides the default key name "adminKey"
      secretAdminKey: "adminToken"
```

By default key of the secret is expected to be `adminKey`

ikatlinsky:
Reopening the issue as the fix was not correct, details here - https://github.com/apache/apisix-helm-chart/pull/568

ikatlinsky:
closing as. https://github.com/apache/apisix-helm-chart/pull/568 was accepted

PR #553, opened 2023-06-02
feat: allow to override admin token for ingress controller

ikatlinsky:
Provides a way to override admin key for ingress controller using secrets, as per proposal https://github.com/apache/apisix-helm-chart/issues/552.

### DIFF
--- a/charts/apisix-ingress-controller/templates/configmap.yaml
+++ b/charts/apisix-ingress-controller/templates/configmap.yaml
@@ -58,7 +58,11 @@ data:
       {{ else }}
       default_cluster_base_url: http://{{ .Values.config.apisix.serviceName }}.{{ .Values.config.apisix.serviceNamespace }}.svc.{{ .Values.clusterDomain }}:{{ .Values.config.apisix.servicePort }}/apisix/admin
       {{- end}}
+      {{- if .Values.config.apisix.secretName }}
+      default_cluster_admin_key: "{{"{{"}}.APISIX_ADMIN_KEY{{"}}"}}"
+      {{- else }}
       default_cluster_admin_key: {{ .Values.config.apisix.adminKey | quote }}
+      {{- end }}
       default_cluster_name: {{ .Values.config.apisix.clusterName | quote }}
 kind: ConfigMap
 metadata:
--- a/charts/apisix-ingress-controller/templates/deployment.yaml
+++ b/charts/apisix-ingress-controller/templates/deployment.yaml
@@ -103,6 +103,13 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.name
+          {{- if .Values.config.apisix.secretName }}
+          - name: APISIX_ADMIN_KEY
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.config.apisix.secretName | quote }}
+                key: {{ include "apisix-ingress-controller.credentials.secretAdminKey" . }}
+          {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
--- a/charts/apisix-ingress-controller/README.md
+++ b/charts/apisix-ingress-controller/README.md
@@ -111,8 +111,10 @@ The same for container level, you need to set:
 | autoscaling.targetCPUUtilizationPercentage | int | `80` |  |
 | autoscaling.version | string | `"v2"` | HPA version, the value is "v2" or "v2beta1", default "v2" |
 | clusterDomain | string | `"cluster.local"` |  |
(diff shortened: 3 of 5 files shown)


ikatlinsky:
@tao12345666333 @tokers can i ask you to take a look at the proposed changes and share your ideas?

tao12345666333 (changes requested):
APISIX Ingress controller behaves differently from APISIX. It cannot obtain configurations in this way.

You can refer to this one https://github.com/apache/apisix-ingress-controller/blob/32c07516b6bf2fe8ff86ecdf65ee04a7afc0c535/test/e2e/suite-chore/config.go#L87-L93

charts/apisix-ingress-controller/templates/configmap.yaml
```diff
@@ -58,7 +58,11 @@ data:
       {{ else }}
       default_cluster_base_url: http://{{ .Values.config.apisix.serviceName }}.{{ .Values.config.apisix.serviceNamespace }}.svc.{{ .Values.clusterDomain }}:{{ .Values.config.apisix.servicePort }}/apisix/admin
       {{- end}}
+      {{- if .Values.config.apisix.existingSecret }}
+      default_cluster_admin_key: "{{"{{"}}.APISIX_ADMIN_KEY{{"}}"}}"
```
lingsamuel, line 62: missing `$` prefix?

ikatlinsky: not really apisix and ingress are using different approaches to access environment variables. Apisix uses a regular approach with $ENV_NAME styles, but ingress is based on go templates, which expects environment variables in {{.ENV_NAME}} format.

ikatlinsky:
https://github.com/apache/apisix-helm-chart/pull/553#pullrequestreview-1482850510

yep, that is why i use {{.ENV}} and not ${{ENV}} notation. We already use it this way internally, it works as expected. I will take a look at your link to see if i missed something.

charts/apisix-ingress-controller/templates/deployment.yaml
```diff
@@ -103,6 +103,13 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.name
+          {{- if .Values.config.apisix.existingSecret }}
+          - name: APISIX_ADMIN_KEY
```
ikatlinsky, line 107: @tao12345666333 i guess your proposal is to try to use `DEFAULT_CLUSTER_ADMIN_KEY` environment variables instead of the `APISIX_ADMIN_KEY`?

tao12345666333: yes

ikatlinsky: in other case i used exactly the same approach as in your tests

ikatlinsky: done

charts/apisix-ingress-controller/templates/configmap.yaml
```diff
@@ -58,7 +58,11 @@ data:
       {{ else }}
       default_cluster_base_url: http://{{ .Values.config.apisix.serviceName }}.{{ .Values.config.apisix.serviceNamespace }}.svc.{{ .Values.clusterDomain }}:{{ .Values.config.apisix.servicePort }}/apisix/admin
       {{- end}}
+      {{- if .Values.config.apisix.existingSecret }}
+      default_cluster_admin_key: "{{"{{"}}.DEFAULT_CLUSTER_ADMIN_KEY{{"}}"}}"
```
tao12345666333, line 62: If ENV is used for configuration, this field is not required in the configuration file.

ikatlinsky: done, thanks

lingsamuel (approved):

JanEgner:
Thanks Ilya! This MR, once approved and released, will allow me to close one of two remaining issues from a security audit on a system I am currently working on. Great work!

ikatlinsky:
@tao12345666333 @AlinsRan anything else is required to progress on these changes?

tao12345666333 (approved):
Thanks

tao12345666333:
If you have a dependency on this PR, you can submit a PR for the release of the Helm chart.
The release process of it is quite flexible.

ikatlinsky:
@tao12345666333 i double checked the solution and it seems like not working, as we still need to provide environment variable in config of ingress controller, please check https://github.com/apache/apisix-helm-chart/pull/568

tao12345666333:
@ikatlinski thanks,  there is indeed a record here.

https://github.com/apache/apisix-ingress-controller/blob/3efd7963a9ed74742639a21e58ce2a14ffe6bb8a/test/e2e/suite-chore/config.go#L38-L40

PR #605, opened 2023-09-05
feat: allow setting ssl/tls ciphers

womfoo:
This will allow finer configuration of ciphers used (ie. disable CBC for LUCKY13 mitigation)


lingsamuel:
it would be better to set empty value as default. If the value is empty, we use default configuration.

womfoo:
worked around the issue with `gateway.tls.sslProtocols: "TLSv1.3"`

PR #624, opened 2023-09-22
feat: support Ingress 1.7.0 etcdserver

lingsamuel:


dickens7:
I think Apisix should use an independent Deployment or directly use the helm chart of apisix for deployment. It is difficult to meet business requirements using only one Deployment.container.

tao12345666333 (approved):

tao12345666333:
Note that this mode is an experimental feature. It aims to remove etcd from the entire architecture. This PR aims to simplify deployment costs by putting them in the same Pod.

Of course, if you have any suggestions, you are welcome to submit a PR to improve it. Thanks

AlinsRan (approved):

VelorumS:
That's confusing. We only wanted to get rid of etcd, but this option also launches another apisix inside the controller's pod.

PR #634, opened 2023-10-06
feat: allow changing service port in the Ingress manifest

indrekj:
Previously it used `service.http.servicePort` in the Ingress manifest. This made it impossible to forward ingress traffic to TLS port. One option was to disable `http` and change `http.servicePort` to `443`, but then regular HTTP couldn't be used.

This commit allows changing Ingress service port using `ingress.servicePort`. It still falls back to `service.http.servicePort` to keep backward compatibility.

We found this useful to enable encryption between AWS ALB and APISIX. This allowed us to use:
```
    alb.ingress.kubernetes.io/backend-protocol: HTTPS
```

### DIFF
--- a/charts/apisix/README.md
+++ b/charts/apisix/README.md
@@ -179,9 +179,10 @@ The command removes all the Kubernetes components associated with the chart and 
 | image.pullPolicy | string | `"IfNotPresent"` | Apache APISIX image pull policy |
 | image.repository | string | `"apache/apisix"` | Apache APISIX image repository |
 | image.tag | string | `"3.5.0-debian"` | Apache APISIX image tag Overrides the image tag whose default is the chart appVersion. |
-| ingress | object | `{"annotations":{},"enabled":false,"hosts":[{"host":"apisix.local","paths":[]}],"tls":[]}` | Using ingress access Apache APISIX service |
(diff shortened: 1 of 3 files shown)


indrekj:
The build should now be green if somebody can approve again. Had some issues with helm-docs before (I thought it was maintained manually).

Revolyssup:
@indrekj If the use case is to allow usage with https then why not do something like `service.tls.servicePort | service.http.servicePort` ?

indrekj:
@Revolyssup Could you elaborate on what exactly you mean? Do you mean using `.Values.service.tls.servicePort | default .Values.service.http.servicePort` in the ingress? I think that would be a backward incompatible change for users that have tls enabled (e.g. for internal traffic) but expect traffic from the load balancer to come from the non-TLS port.

Revolyssup:
> @Revolyssup Could you elaborate on what exactly you mean? Do you mean using `.Values.service.tls.servicePort | default .Values.service.http.servicePort` in the ingress? I think that would be a backward incompatible change for users that have tls enabled (e.g. for internal traffic) but expect traffic from the load balancer to come from the non-TLS port.

Hmm, you're right.

Revolyssup (approved):

lingsamuel (approved):
LGTM

PR #635, opened 2023-10-06
feat: allow adding extra containers to apisix deployment

indrekj:
For our use case, we want to deploy our authentication container next to the APISIX so there's minimal latency when using `forward-auth` plugin.


indrekj:
The build should now be green if somebody can approve again. Had some issues with helm-docs before (I thought it was maintained manually).

Revolyssup (approved):

shreemaan-abhishek (approved):

tao12345666333 (approved):

PR #636, opened 2023-10-09
feat: allow deploying additional kubernetes resources

indrekj:
The `extraDeploy` option is something that is available in many helm charts (expecially bitnami charts). It is useful for deploying extra resources with the release. E.g. with Apisix you could deploy ApisixGlobalRule.

This commit adds a very simple version of the extraDeploy.

E.g. Using as a string:

```
extraDeploy:
  - |
    apiVersion: apisix.apache.org/v2
    kind: ApisixGlobalRule
    metadata:
      name: global
    spec:
      plugins:
        - name: response-rewrite
          enable: true
          config:
            headers:
              remove:
                - Server
```

can also use it as a map:

```
  - apiVersion: apisix.apache.org/v2
    kind: ApisixGlobalRule
    metadata:
      name: global
    spec:
      plugins:
        - name: response-rewrite
          enable: true
          config:
            headers:
              remove:
                - Server
```


Revolyssup (approved):

indrekj:
Rebased against dev to fix merge conflict. Workflow needs to be approved again.

PR #637, opened 2023-10-09
feat: allow adjusting nginx keepalive timeout

indrekj:
Depending on the setup, it can be useful to change this value. For example, if using load balancer which already has 60 second timeout, then this timeout should be larger to avoid ungraceful TCP connection closes.

This is the same as `upstream-keepalive-timeout` in ingress nginx controller.

E.g. From AWS [docs][1]:

> We also recommend that you configure the idle timeout of your
application to be larger than the idle timeout configured for the load balancer. Otherwise, if the application closes the TCP connection to the load balancer ungracefully, the load balancer might send a request to the application before it receives the packet indicating that the connection is closed. If this is the case, the server refuses the request from the load balancer and then the load balancer sends an HTTP 502 Bad Gateway error to the application.

[1]: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout


Revolyssup (approved):
LGTM

lingsamuel (approved):
Thanks!

PR #638, opened 2023-10-11
chore: update ApisixRoute crd paths pattern rule

g-sponda:
## Description
The current paths item pattern regex doesn't accept to have wildcards in the middle of the path, meanwhile it's possible to create this Apisix route calling the Apisix API or through Apisix Dashboard.

My currently issue is that I need to match the following URLs:
```
/my/api/country/br/cities
/my/api/country/us/cities
```

But with the current pattern set on the CRD, I'm not able to create the spec.http.match.paths `/my/api/country/*/cities`  through the CRD.


lingsamuel (approved):

tao12345666333 (approved):
LGTM

PR #643, opened 2023-10-15
feat: auto populate ingress_publish_service when using etcdserver

zhuoyang:
When using the new composite architecture, ingress resources will throw out this error due to the service of the gateway not being set
```
error	ingress/ingress.go:502	failed to get APISIX gateway external IPs	{"error": "service \"\" not found"}
```

This PR set the `ingress_publish_service` to the default gateway service when etcdserver is enabled and the `ingressPublishService` values is empty


AlinsRan (approved):

tao12345666333 (approved):
LGTM

PR #646, opened 2023-10-18
fix: plugin_metadata_cm path

zhuoyang:
It is supposed to be on root level instead of under kubernetes
https://github.com/apache/apisix-ingress-controller/blob/0736bba3f3bda0d47620ef1c6414f73d5b2056e5/pkg/config/config.go#L94


lingsamuel (approved):

AlinsRan (approved):

indrekj:
Awesome. I was struggling with this issue for half a day today without noticing there is an open PR already. Could this be also applied to the `dev` branch?

Gallardot (approved):

Gallardot:
> Awesome. I was struggling with this issue for half a day today without noticing there is an open PR already. Could this be also applied to the `dev` branch?

Yes, welcome to create a PR.

indrekj:
> Yes, welcome to create a PR.

Ok, done: https://github.com/apache/apisix-helm-chart/pull/648

Gallardot:
> EDIT: nvm, we just need to bump the ingress controller chart version when it is released. No need to duplicate the code.

Yes, you are right.I make a mistask.

PR #648, opened 2023-10-20
fix: plugin_metadata_cm path

indrekj:
Exactly the same thing as #646 but for the dev branch (v2.x chart)


indrekj:
Oh actually maybe this doesn't make sense. Dev branch is for the `apisix` chart right? Not for `apisix-ingress-controller` chart? So we should rather bump the `apisix-ingress-controller` chart version when it is released and not port this over? I'll close it. Was a bit too hasty.

PR #654, opened 2023-10-27
Add podLabels to _pod.tpl

gamer22026:
Allow adding additional labels to pod


Gallardot (commented):
Unable to find the definition of `podLabels` in the project

Issue #681, opened 2023-12-08
TZ and extraEnvVars in apisix-ingress-controller helm chart


papdaniel:
Apisix helm chart has TZ and extra env vars config, these would be useful in apisix-ingress-controller too. 
Currently it runs in UTC+8, we would like to set it to another timezone.

alangilmor:
volumes & volumeMounts also, please

PR #682, opened 2023-12-08
feat: timezone and extra env vars configurable in apisix ingress controller

papdaniel:
Apisix helm chart has TZ and extra env vars config, these would be useful in apisix-ingress-controller too.
Currently it runs in UTC+8, we would like to set it to another timezone.
issue: https://github.com/apache/apisix-helm-chart/issues/681

### DIFF
--- a/charts/apisix-ingress-controller/README.md
+++ b/charts/apisix-ingress-controller/README.md
@@ -143,6 +143,7 @@ The same for container level, you need to set:
 | config.logLevel | string | `"info"` | the error log level, default is info, optional values are: debug, info, warn, error, panic, fatal |
 | config.logOutput | string | `"stderr"` | the output file path of error log, default is stderr, when the file path is "stderr" or "stdout", logs are marshalled plainly, which is more readable for human; otherwise logs are marshalled in JSON format, which can be parsed by programs easily. |
 | config.pluginMetadataCM | string | `""` | Pluginmetadata in APISIX can be controlled through ConfigMap. default is "" |
+| extraEnvVars | list | `[]` | extraEnvVars An array to add extra env vars e.g: extraEnvVars:   - name: FOO     value: "bar"   - name: FOO2     valueFrom:       secretKeyRef:         name: SECRET_NAME         key: KEY |
 | fullnameOverride | string | `""` |  |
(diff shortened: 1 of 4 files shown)


charts/apisix-ingress-controller/README.md
```diff
@@ -185,6 +186,7 @@ The same for container level, you need to set:
 | serviceMonitor.annotations | object | `{}` | @param serviceMonitor.annotations ServiceMonitor annotations |
 | serviceMonitor.labels | object | `{}` | @param serviceMonitor.labels ServiceMonitor extra labels |
 | serviceMonitor.metricRelabelings | object | `{}` | @param serviceMonitor.metricRelabelings MetricRelabelConfigs to apply to samples before ingestion. ref: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs |
+| timezone | string | `""` | timezone is the timezone where apisix uses. For example: "UTC" or "Asia/Shanghai" This value will be set on apisix-ingress-controller container's environment variable TZ. You may need to set the timezone to be consistent with your local time zone, otherwise the apisix-ingress-controller's logs may be used to retrieve event in wrong timezone. |
```
zll600, line 189:
```suggestion
| timezone | string | `""` | timezone is the timezone where apisix-ingress-controller uses. For example: "UTC" or "Asia/Shanghai" This value will be set on apisix-ingress-controller container's environment variable TZ. You may need to set the timezone to be consistent with your local time zone, otherwise the apisix-ingress-controller's logs may be used to retrieve event in wrong timezone. |
```

charts/apisix-ingress-controller/values.yaml
```diff
@@ -183,6 +183,25 @@ affinity: {}
 # ref: https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/#spread-constraints-for-pods
 topologySpreadConstraints: []
 
+# -- timezone is the timezone where apisix uses.
```
zll600, line 186:
```suggestion
# -- timezone is the timezone where apisix-ingress-controller uses.
```

zll600:
@Revolyssup @AlinsRan Could you help review this?

PR #685, opened 2023-12-17
Fix already defined hostNetwork when it is true

jaysonsantos:


zll600:
@Revolyssup Can you help review this PR?

PR #692, opened 2024-01-05
feat: Support config merge from Kubernetes Secret

achetronic:
Hello,

We have detected that some of the credentials regarding configuration are hardcoded on `values.yaml`. We needed to fix this for our company as we use ExternalSecrets operator to provide the credentials dynamically to the Kubernetes clusters. 

We have patched the Helm chart to be able to merge some config coming from a Kubernetes Secret with the main config coming from the dashboard's configmap. This way we cover all the potential secret fields across config (etc, authentication users, etc)

For merging we are using the official `yq` image so we expect you like it

WDYT about changes?


achetronic:
Let's close this PR as null response given

PR #698, opened 2024-01-15
fix: allow configuring apiVersion in Ingress Class resource

Revolyssup:
Fixes https://github.com/apache/apisix-ingress-controller/issues/2133

### DIFF
--- a/charts/apisix-ingress-controller/templates/ingress-class.yaml
+++ b/charts/apisix-ingress-controller/templates/ingress-class.yaml
@@ -1,5 +1,7 @@
-{{ if (eq .Values.config.kubernetes.ingressVersion "") }}
+{{ if or (eq .Values.config.kubernetes.ingressVersion "") (eq .Values.config.kubernetes.ingressVersion "networking/v1")}}
 apiVersion: networking.k8s.io/v1
+{{- else if (eq .Values.config.kubernetes.ingressVersion "networking/v1beta1")}}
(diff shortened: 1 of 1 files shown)


Gallardot (approved):

AlinsRan (approved):

charts/apisix-ingress-controller/templates/ingress-class.yaml
```diff
@@ -1,4 +1,8 @@
+{{ if (eq .Values.config.kubernetes.ingressVersion "") }}
 apiVersion: networking.k8s.io/v1
+{{- else }}
+apiVersion: {{ .Values.config.kubernetes.ingressVersion }}
```
AlinsRan, line 4: https://github.com/apache/apisix-ingress-controller/blob/master/conf/config-default.yaml#L72-L74

Are you sure you can use it directly?
`networking/v1beta1 != networking.k8s.io/v1beta1`

AlinsRan (changes requested):

charts/apisix-ingress-controller/templates/ingress-class.yaml
Revolyssup: nice catch

Revolyssup:
@AlinsRan Can you review again? I made the changes/

AlinsRan (approved):

PR #706, opened 2024-01-17
feat: add serviceMonitor for apisix-gateway

mstefany:
Metrics' port & URL is defined [here](https://github.com/apache/apisix-helm-chart/blob/f14f58dc7fbe8a963b346ab322a93c662f191b3a/charts/apisix-ingress-controller/templates/apisix-configmap.yaml#L170-L171) (port 9091 & URL `/apisix/prometheus/metrics`).

And it doesn't match what's in the [serviceMonitor](https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix-ingress-controller/templates/servicemonitor.yaml#L33). TargetPort should be `prometheus`, a.k.a. 9091, and path is not specified which defaults to `/metrics` which is incorrect.

Fact that it's not working in default configuration is visible in my logs:
```
apisix-ingress-controller-6b4d5cc778-f582v apisix 10.254.82.207 - - [17/Jan/2024:11:00:22 +0000] 10.254.83.238:9080 "GET /metrics HTTP/1.1" 404 47 0.000 "-" "Prometheus/2.48.1" - - - "http://10.254.83.238:9080"
```

### DIFF
--- a/charts/apisix-ingress-controller/README.md
+++ b/charts/apisix-ingress-controller/README.md
@@ -154,6 +154,11 @@ The same for container level, you need to set:
 | gateway.nginx.workerRlimitNofile | string | `"20480"` | Nginx workerRlimitNoFile |
 | gateway.resources | object | `{}` |  |
 | gateway.securityContext | object | `{}` |  |
+| gateway.serviceMonitor | object | `{"annotations":{},"enabled":false,"interval":"15s","labels":{},"metricRelabelings":{},"namespace":"monitoring","path":"/apisix/prometheus/metrics"}` | Create ServiceMonitor object for apisix gateway for Prometheus operator Requires Prometheus operator v0.38.0 or higher. |
+| gateway.serviceMonitor.annotations | object | `{}` | @param serviceMonitor.annotations ServiceMonitor annotations |
+| gateway.serviceMonitor.labels | object | `{}` | @param serviceMonitor.labels ServiceMonitor extra labels |
+| gateway.serviceMonitor.metricRelabelings | object | `{}` | @param serviceMonitor.metricRelabelings MetricRelabelConfigs to apply to samples before ingestion. ref: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs |
+| gateway.serviceMonitor.path | string | `"/apisix/prometheus/metrics"` | @param serviceMonitor.path ServiceMonitor scrape path |
 | gateway.tls.additionalContainerPorts | list | `[]` | Support multiple https ports, See [Configuration](https://github.com/apache/apisix/blob/0bc65ea9acd726f79f80ae0abd8f50b7eb172e3d/conf/config-default.yaml#L99) |
 | gateway.tls.certCAFilename | string | `""` | Filename be used in the gateway.tls.existingCASecret |
 | gateway.tls.containerPort | int | `9443` |  |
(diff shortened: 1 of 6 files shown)


charts/apisix-ingress-controller/templates/servicemonitor.yaml
```diff
@@ -30,8 +30,9 @@ metadata:
   {{- end }}
 spec:
   endpoints:
-  - targetPort: http
+  - targetPort: prometheus
     scheme: http
+    path: /apisix/prometheus/metrics
```
zll600, line 35: Do we need to add a new value `serviceMonitor`, like https://github.com/apache/apisix-helm-chart/blob/f14f58dc7fbe8a963b346ab322a93c662f191b3a/charts/apisix/values.yaml#L571

mstefany: I can make the value configurable, and fix it in both places.

mstefany: @zll600 I added `serviceMonitor.path` as configurable option, please review

zll600: LGTM, Thank you

mstefany:
Please, is it possible to have this reviewed and approved?

zll600:
ping @AlinsRan @Revolyssup

zll600:
ping @AlinsRan @Revolyssup

flassagn:
Guys any feedback about the review ? We would like to enable the ingress controller monitoring ! :pray:

Revolyssup:
> Guys any feedback about the review ? We would like to enable the ingress controller monitoring ! 🙏

I will take a look at this tomorrow and review. Thanks for reminding.

Revolyssup (approved):

mstefany:
I pushed the updated version, so that linter will be happy this time.

flassagn:
Hi @Revolyssup,

charts/apisix-ingress-controller/templates/servicemonitor.yaml
```diff
@@ -30,8 +30,9 @@ metadata:
   {{- end }}
 spec:
   endpoints:
-  - targetPort: http
+  - targetPort: prometheus
```
AlinsRan, line 33: The original HTTP should be the metrics of ingress. 
You should add servicemonitor instead of replacing it.

Why replace it with the metrics of the data plane? https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/templates/service-monitor.yaml#L41

How did you deploy it?

mstefany: Yeah, you're right. I will rewrite the PR. To scrape both apisix-ingress-controller and apisix (gateway) properly.

mstefany:
@AlinsRan @Revolyssup I updated the PR:

- because both `apisix-ingress-controller` and `apisix` containers have ports named `http`, original `ServiceMonitor` would scrape both containers (if `apisix-gateway` is enabled), and cause error every 15 seconds, because there are no metrics on http port 9080 of gateway at `/metrics` URL. For this reason I am renaming `http` port of `apisix-gateway-controller` to `http-controller`.
- enabling `ServiceMonitor` of gateway is now done separately in `.Values.gateway.serviceMonitor`, and it also provisions separate `ServiceMonitor` object to scrape such metrics

Please, review the code if this approach is acceptable. I have updated `README.md` with `helm-docs` so there shouldn't be linter errors.

flassagn:
Hi Guys, 

@AlinsRan @Revolyssup could you review the PR shared by @mstefany ? :pray:  

Thank you for your time and I wish you a good day

PR #712, opened 2024-01-24
feat: allow to enable/disable some plugins

mstefany:
List of plugins for gateway (apisix container) is currently configured statically in the ConfigMap and there is no way to customize the list. That means that e.g. `zipkin` plugin for observability is always enabled, but `opentelemetry` is not and it is not possible to enabled it (easily).

This just a proposal, how it could be done in one way.

I was also thinking about:

```
gateway:
  pluginsOverride:
    - plugin1
    - plugin2
    - ...
```

Or instead of adjusting ConfigMap template to allow for swappping it with `existingConfigMap`, e.g.:

```
gateway:
  existingConfigMap: "my-apisix-configmap"
```

Which would cause to not render `apisix-configmap.yaml` at all and instead switch `configuration` volume to provided custom ConfigMap.

Feedback is welcome.

### DIFF
--- a/charts/apisix-ingress-controller/README.md
+++ b/charts/apisix-ingress-controller/README.md
@@ -152,6 +152,9 @@ The same for container level, you need to set:
 | gateway.nginx.workerConnections | string | `"10620"` | Nginx worker connections |
 | gateway.nginx.workerProcesses | string | `"auto"` | Nginx worker processes |
 | gateway.nginx.workerRlimitNofile | string | `"20480"` | Nginx workerRlimitNoFile |
+| gateway.plugins.opentelemetry.enabled | bool | `false` |  |
+| gateway.plugins.skywalking.enabled | bool | `false` |  |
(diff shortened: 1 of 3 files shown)


mstefany:
Any possibility for a review and/or discussion, please?
It would be actually useful to control both list of enabled plugins and `plugin_attrs` in the ConfigMap.

One new option that could be useful is:
```
gateway:
  plugin_attrs:
    opentelemetry:
      ...
  plugins:
    enable:
      - plugin1
      - plugin2
    disable:
      - plugin3
      - plugin4
```

gateway.plugins.enable would add something to the list of default plugins, while disable would remove it. My concern here is - if plugins need to be specified in the list by their priority or if order just doesn't matter.

Revolyssup:
@mstefany Sorry for the late reply. I got busy. Now I am going over all your proposals and PRs one by one

Revolyssup:
@mstefany I will approve this PR but the proposal you have, can you create a separate issue and add details in the description so that other maintainers can take a look at the proposal?

Issue #653, opened 2023-10-26
Apisix 2.3.0 missing pdb fix


jaysonsantos:
Hi there folks, on version 2.3.0 the fix from #508 
would it be possible to cut a new release with that? otherwise all clusters with 1.25> would break

zkl94:
this is very much needed. That commit was merged long time ago but it is still not in latest release

michbeck100:
I think the changes from master are part of the 1.x.y chart releases. I don't understand the versioning here, though. I was also missing a merged PR in the latest chart. But i found it in 1.7.1.

Maybe someone in the team can explain what the differences between 2.x and 1.x are.

zll600:
There is an explanation. https://github.com/apache/apisix-helm-chart/issues/656

Baoyuantop:
The PDB fix referenced in this issue has been included in subsequent chart releases. The chart has since iterated to v2.13.0, which contains this fix. This issue has also had no activity for over a year.

Closing as resolved. If you still encounter this issue, please feel free to reopen.

Issue #680, opened 2023-12-07
apisix-2.4.0 chart uses deprecated  poddisruptionbudget v1beta1


atamgp:
[deprecation-guide](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125)

poddisruptionbudget v1beta1 is deprecated, please update in the next chart version.
for now I have disabled it.

zll600:
Accoding this https://github.com/apache/apisix-helm-chart/issues/656.
> the dev branch is only used for 2.x. and master branch is used for other.

the issue on master was fixed by https://github.com/apache/apisix-helm-chart/pull/508. So we will create a same fix for dev which can help fix the 2.x version.

Baoyuantop:
The PodDisruptionBudget template has been updated to use `semverCompare` to automatically select the appropriate API version (`policy/v1` or `policy/v1beta1`) based on the Kubernetes cluster version. This compatibility issue has been resolved.

Closing as fixed. If you still encounter this issue, please feel free to reopen.

PR #508, opened 2023-03-12
PDB updated to Kubernetes 1.25

lpiob:
Why:
* Chart could not be installed in Kubernetes 1.25
* The PodDisruptionBudget used apiVersion that is deprecated before and unavailable in 1.25.

How:
* PodDisruptionBudget apiVersion is updated to match the Kubernetes version. No spec adjustments needed.
* Changes are applied also to k8s versions in which old api versions were deprecated.

Reference:
* https://kubernetes.io/docs/reference/using-api/deprecation-guide/#poddisruptionbudget-v125
   
I have tested the changes on a 1.25 cluster and reviewed templates visually for older version.

PR #713, opened 2024-01-29
fix: update pdb version to k8s 1.25

zll600:
fix: https://github.com/apache/apisix-helm-chart/issues/653, https://github.com/apache/apisix-helm-chart/issues/680

this pr only copy code from https://github.com/apache/apisix-helm-chart/pull/508

### DIFF
--- a/charts/apisix/templates/pdb.yaml
+++ b/charts/apisix/templates/pdb.yaml
@@ -18,8 +18,9 @@
 {{ if semverCompare "<1.21-0" .Capabilities.KubeVersion.Version -}}
 apiVersion: policy/v1beta1
 {{- else -}}
-kind: PodDisruptionBudget
+apiVersion: policy/v1
 {{- end }}
+kind: PodDisruptionBudget
 metadata:
   name: {{ include "apisix.fullname" . }}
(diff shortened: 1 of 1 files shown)


charts/apisix/templates/pdb.yaml
```diff
@@ -15,8 +15,11 @@
 # limitations under the License.
 
 {{- if (.Values.podDisruptionBudget.enabled) }}
+{{ if semverCompare "<1.21-0" .Capabilities.KubeVersion.Version -}}
 apiVersion: policy/v1beta1
+{{- else -}}
 kind: PodDisruptionBudget
+{{- end }}
```
Revolyssup, line 22: Did you mean to specify apiVersion policy/v1 here? I dont think the logic is quite right.

Revolyssup (changes requested):

charts/apisix/templates/pdb.yaml
Revolyssup: https://github.com/apache/apisix-helm-chart/pull/719#pullrequestreview-1948146765

zll600: Sorry, I will update it.

Issue #705, opened 2024-01-17
Add helm values to use an external configmap on APISIX standalone mode


tkyoo:
When using APISIX standalone mode as a subchart, 'apisix.yaml' configmap is always re-initialized when upgrading the parent chart.

I think using an external configmap as the "apisix.yaml" is a simple solution for solving them!

Example: (variable names and paths are very sensitive part of the project, so this code is just a sample!)

apisix-config-configmap.yaml:
```yaml
{{- if and (eq .Values.deployment.mode "standalone") (eq (len .Values.apisix.standalone.extraConfigMap) 0 ) }}
kind: ConfigMap
apiVersion: v1
metadata:
  name: apisix.yaml
data:
  apisix.yaml: |
    routes:
    -
      uri: /hi
      upstream:
        nodes:
          "127.0.0.1:1980": 1
        type: roundrobin
    #END
{{- end }}
```

_pod.tpl:
```yaml
  volumes:
    {{- if eq .Values.deployment.mode "standalone" }}
    - configMap:
        name: {{ .Values.apisix.standalone.extraConfigMap | default "apisix.yaml" }}
      name: apisix-admin
    {{- end }}
```

values.yaml:
```yaml
apisix:
  ...
  standalone:
    extraConfigMap: "apisix-yaml"
```

Maybe this feature will also be very helpful for other APISIX users :)

Gallardot:
It's a meaningful suggestion, but few users would do this, especially in a k8s cluster. Supporting such a feature would require users to ensure correct configuration on their own, which could introduce risks.

jompu:
> It's a meaningful suggestion, but few users would do this, especially in a k8s cluster. Supporting such a feature would require users to ensure correct configuration on their own, which could introduce risks.

Isn't it the idea on standalone mode, that you can edit this file? Now it is not possible to use standalone mode on kubernetes (with this helm chart) because you cannot actually add any configuration there.  Unless if you are fine with editing the configmap straight on the cluster.

For example, I would like to have this configuration stored in git repository which I use for deployment. But currently it is impossible.

jompu:
I have created a pull request with related changes. The pull request allows to override content of the apisix.yaml ConfigMap or to use existing ConfigMap.

Meanwhile you can use either branch `dev` or `more-standalone-config-options` from https://github.com/jompu/apisix-helm-chart to install modified version of the APISIX 2.6.0 chart.

Baoyuantop:
[...]

PR #717, opened 2024-02-02
More options for standalone configuration

jompu:
Related to issue: #705

Allow to set standalone configuration with chart values or by using existing ConfigMap.

### DIFF
--- a/charts/apisix/templates/apisix-config-cm.yml
+++ b/charts/apisix/templates/apisix-config-cm.yml
@@ -21,12 +21,6 @@ metadata:
   name: apisix.yaml
 data:
   apisix.yaml: |
-    routes:
-    -
-      uri: /hi
-      upstream:
-        nodes:
-          "127.0.0.1:1980": 1
-        type: roundrobin
+    {{- .Values.deployment.standaloneConfig | nindent 4 }}
     #END
 {{- end }}
--- a/charts/apisix/templates/apisix-config-cm.yml
+++ b/charts/apisix/templates/apisix-config-cm.yml
@@ -21,6 +21,6 @@ metadata:
   name: apisix.yaml
 data:
   apisix.yaml: |
-    {{- .Values.deployment.standaloneConfig | nindent 4 }}
+    {{- .Values.apisix.deployment.standaloneConfig | nindent 4 }}
     #END
 {{- end }}
--- a/charts/apisix/templates/apisix-config-cm.yml
+++ b/charts/apisix/templates/apisix-config-cm.yml
@@ -14,13 +14,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-{{- if eq .Values.apisix.deployment.mode "standalone" }}
+{{- if and (eq .Values.apisix.deployment.mode "standalone") (not .Values.apisix.deployment.standalone.existingConfigMap) }}
 kind: ConfigMap
 apiVersion: v1
 metadata:
   name: apisix.yaml
 data:
   apisix.yaml: |
-    {{- .Values.apisix.deployment.standaloneConfig | nindent 4 }}
+    {{- .Values.apisix.deployment.standalone.config | nindent 4 }}
     #END
 {{- end }}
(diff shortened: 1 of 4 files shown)


charts/apisix/templates/apisix-config-cm.yml
exxbrain, line 1: What is this config map for?

charts/apisix/values.yaml
```diff
@@ -305,6 +305,24 @@ apisix:
...
+        routes:
+        -
+          uri: /hi
+          upstream:
+            nodes:
+              "127.0.0.1:1980": 1
+            type: roundrobin
+      # -- Specifies the name of the ConfigMap that contains the rule configurations.
+      # The configuration must be set to the key named `apisix.yaml` in the configmap.
+      existingConfigMap: ""
```
exxbrain, line 324: Is this really needed?

jompu: It is for this subchart issue: https://github.com/apache/apisix-helm-chart/issues/705

Though I have not tested if this is really needed for that use case. But I think there are many reasons that some one would like to use a configmap that is created outside of this chart. For example related to devops automation, where the configuration is defined/updated separate to the chart installation.

espizo: We need this as well. The current Helm chart adds a apisix.yaml ConfigMap when it is installed in standalone mode. This ConfigMap contains a /hi route that is impossible to modify or remove. We need the ability to specify our own rule config instead of this example config.

One workaround would maybe be to "overwrite" the example config mount using `extraVolumes` and `extraVolumeMounts` but I much rather prefer this solution.

merriadoc:
This pull request will be wonderfull ! :-)

kedare:
Something that could be interesting to add, we should name the configmap dynamically with the name of the relase (like most others k8s objects I think right now)

As of today we cannot have multiple releases of apisix in a single namespace because of this (conflict on the fixed `apisix.yaml` configmap name)

jompu:
The 2.x development has changed from dev to master since I created this, so I have updated the branch to match latest master. And this pull request to merge into master.

jompu:
@Gallardot Maybe you can help to get this forward also.

The chart is not usable in standalone mode without this.

bruhsb:
hey folks o/

I'm facing the same issue, I needed to use `strategicMergePatches` in `helmfile` to rewrite the volume `apisix-admin` to use a custom configmap T_T

Can I do something to help this PR to move forward? :)

ranjanashish:
@Gallardot @exxbrain Can one of you help move this forward please?

As others have noted, in its current state, this Helm chart is unusable for standalone mode. This is particularly surprising because even the [FAQs](https://github.com/apache/apisix-helm-chart/blob/c52d29d3fbc5bfbab506581fccb47bbe2f6a3d2f/docs/en/latest/FAQ.md#how-to-install-apache-apisix-running-in-standalone-mode) has details on how to deploy APISIX in standalone mode. And yes, the command used in the FAQ will deploy APISIX in standalone mode but won't let you configure routes or anything - essentially it will deploy a dummy APISIX that is not usable.

j4rj4r:
Hello ! 

This pull request is essential for ensuring a fully functional and easily deployable standalone mode.
As it stands, the current Helm chart configuration isn’t practical for automation or GitOps.

thanks,

jompu:
I have moved to use Bitnami's APISIX helm chart, which is much more versatile and allows also standalone configuration.

TheShubhendra:
I am facing the same issue. As a quick work around after deploying the apisix, I am updating the `apisix.yaml` ConfigMap from another yaml file

paalgyula:
@exxbrain what is the status on your review? It would be really helpful for us too!

jompu:
> @exxbrain @espizo Would it be possible to move forward with the review? As seen in the comments on this PR, this feature would be appreciated by many.

Especially because Bitnami chart is not alternative option anymore.

adir-cynomi:
Hello @nic-chen @AlinsRan 
Should we expect a new helm chart version to be released containing these new changes?
The latest (2.12.4) doesn't contain them so still can't be used.

PR #718, opened 2024-02-02
Make the path type configurable in ingress rules.

jompu:
There were some problems at least with ALB ingress controller and APISIX gateway, when the path type is ImplementationSpecific. Setting path type to Prefix fixed the issue.


kedare:
+1: This is blocking for us, we can't use APISIX behind the ALB as it's only routing to strictly `/` and not `/*`

kedare:
(I would also make it configurable on the admin ingress ?)

kedare:
@Gallardot @AlinsRan Hello, would you be able to take a quick look at this one ? It should be quite fast to review and merge :)

Thanks

Gallardot (approved):
LGTM

Gallardot (changes requested):
I'm sorry, I didn't notice CI's status.

We should make the CI happy. @jompu

jompu:
> I'm sorry, I didn't notice CI's status.
> 
> We should make the CI happy. @jompu

The 2.x development has changed from dev to master since I created this, so I have updated the branch to match latest master. And this pull request to merge into master.

PR #722, opened 2024-02-12
Add support for custom ports

### LINKED ISSUE #720: How to expose prometheus endpoint without ServiceMonitor ?
Hello.

I am trying to make Prometheus scrape the APISIX endpoint without success.

I don't want to use the `ServiceMonitor` because it's not managed directly in the cluster, so basically I just need to add the annotations (done via `apisix.podAnnotations`) and then expose the port, which is the blocking part.

It looks like there is no way to expose more ports from the pods ?
I tried the `gateway.http.additionalContainerPorts` but then it get in conflicts as it's also setting up APISIX to listen on this port.

I'm quite surprised to use `ServiceMonitor` that is not really standard nor useful for this kind of case, as all what it is setting up can be setup using pod annotations and would work even without dependencies installed (here I can't use the ServiceMonitor because the CRD is not supported)

Thank you.

kedare:
This PR add supports for custom ports without impacting the apisix configuration by itself.

It is useful to expose ports that may be exposed from specific apisix configuration like Prometheus.

This makes possible to monitor APISIX with Prometheus using only standard Kubernetes annotations and without requiring ServiceMonitor CRD support (it can also likely solve others usercases)

Fixed #720

### DIFF
--- a/charts/apisix/Chart.yaml
+++ b/charts/apisix/Chart.yaml
@@ -31,7 +31,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.11.0
+version: 1.11.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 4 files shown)


kedare:
Any update on this ? Can we help more ? Looks like I am missing some access to be able to solve the conflicts

ralbon:
@ikatlinsky Please, can you have a look ?

notsag:
Please, @AlinsRan @Gallardot, when you have time could you review this PR.
Thanks

PR #740, opened 2024-03-21
feat: Autogenerate random API Key with helm install if not passed

Revolyssup:
This PR is part of the proposal which removes hardcoding of sensitive API keys by autogenerating them at either application level(APISIX) or deployment level(like helm chart.)


charts/apisix/templates/configmap.yaml
```diff
@@ -326,6 +326,8 @@ data:
           - name: "admin"
             {{- if .Values.apisix.admin.credentials.secretName }}
             key: ${{"{{"}}APISIX_ADMIN_KEY{{"}}"}}
+            {{- else if eq .Values.apisix.admin.credentials.admin "" }}
+            key: {{ randAlphaNum 32 }}
```
AlinsRan, line 330: How do users know the key? You need to add a prompt output.

Perhaps managing admin-key through a secret would be more appropriate.
What do you think？

Revolyssup: Since this is in the confimap. Users can look at the configmap using kubectl to look at the key. Is there an issue there?

starsz: I think we can write how to get the key in the docs.

Revolyssup: I have added that in the README against this particular key. https://github.com/apache/apisix-helm-chart/pull/740/files#diff-3169e4e147b847e17edc927d65d4aa9cb5db1a436d62acf059c166e898f9f3fdR46

AlinsRan: You can add a prompt here.
https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/templates/NOTES.txt

AlinsRan: Is it regenerated when upgrading the version?

PR #757, opened 2024-05-22
bugfix values.yaml dns config

hamishliu:
Fixed helm deployment package configuration error, the original dns configuration item does not take effect


Revolyssup:
@hamishliu Can you describe more clearly the purpose of this PR and the problem it solves?

PR #787, opened 2024-11-04
Allow nodePort & servicePort also to be configured in additionalContainerPorts

pavankumar-go:
currently for configuring gateway service only the port parameter is used to populate TargetPort and ServicePort. which is quite limiting in certain conditions
```
    additionalContainerPorts:
      - port: 9081
```

With added changes:
```
    additionalContainerPorts:
      - port: 8443
        servicePort: 80    # optional, if not set uses `.port` value, hence it does not cause breaking changes
        nodePort: 30303 # optional
```


luismiguelsaez-steercrm:
Seems similar to the one I opened some time ago: https://github.com/apache/apisix-helm-chart/pull/782

Anyway, it looks like no one is taking any of them into account :(

PR #793, opened 2024-12-04
feat: make prometheus shared dict configurable

### LINKED ISSUE #769: Nginx errors in metric. Increase lua_shared_dict prometheus-metrics value in helm.
Hello, we are facing an issue described at https://github.com/apache/apisix/issues/7349. Please advise on how to add the nginx parameter during deployment via the Helm chart:

```
http:
  lua_shared_dict:
    prometheus-metrics: 10m
```

After adding via customLuaSharedDicts:, I encountered the following error:

```
[emerg] 1#1: the size 536870912 of shared memory zone "prometheus-metrics" conflicts with already declared size 10485760 in /usr/local/apisix/conf/nginx.conf:100

```
The reason for adding the parameter is increasing errors in the apisix_nginx_metric_errors_total metric:

image

It has also been noticed that when a certain number of errors is reached, the metrics stop displaying correctly.

acuteaura:
ref: https://github.com/apache/apisix/blob/ee6a04cad1bab24ae0305a3cd66bb4e88812edb7/conf/config.yaml.example#L166

fixes #769

### DIFF
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -448,7 +448,7 @@ apisix:
       # - name: bar
       #   size: 1m
 
-    # Set the size of prometheus-metrics shared dict
+    # -- Set the size of prometheus-metrics shared dictionary
     prometheusSharedDictSize: 15m
 
   discovery:
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -448,6 +448,9 @@ apisix:
       # - name: bar
       #   size: 1m
 
+    # Set the size of prometheus-metrics shared dict
+    prometheusSharedDictSize: 15m
+
   discovery:
     # -- Enable or disable Apache APISIX integration service discovery
     enabled: false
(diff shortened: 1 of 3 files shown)


charts/apisix/values.yaml
```diff
@@ -448,6 +448,9 @@ apisix:
       # - name: bar
       #   size: 1m
 
+    # Set the size of prometheus-metrics shared dict
+    prometheusSharedDictSize: 15m
```
bzp2010, line 452: 🤔 Is it a good idea to use a separate orphaned key?

Wouldn't it be better to implement it as a map and allow the original shared dict name used by APISIX to be passed in directly for overriding. e.g.

```yaml
luaSharedDicts:
  - name: prometheus-metrics
    size: 100m
```
If you think this is unclear, perhaps it could also be called `builtinLuaSharedDicts`. Or, make it configurable by `customLuaSharedDicts`, although not intuitive.

BTW, if you update the values template, please update the documentation along with it.

acuteaura: it is very specific (in the otherwise unused meta key) in the upstream config, and the same mechanism doesn't seem to be used anywhere else, so I opted for a single specific key

shihuncl: 添加了上述配置启动的时候报错如下：2025/01/10 20:06:58 [emerg] 1#1: the size 52428800 of shared memory zone "prometheus-metrics" conflicts with already declared size 10485760 in /usr/local/apisix/conf/nginx.conf:96
nginx: [emerg] the size 52428800 of shared memory zone "prometheus-metrics" conflicts with already declared size 10485760 in /usr/local/apisix/conf/nginx.conf:96，怎么去解决这个问题？

Issue #667, opened 2023-11-16
Create a new proxy_cache zone using helm chart


chilcano:
Hi there!
Is there a way to create a new `proxy_cache` zone through existing helm chart?.

This is existing configuration used by helm chart:
* helm chart: https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/templates/configmap.yaml
* config.yaml: https://github.com/apache/apisix/blob/master/conf/config-default.yaml

If so not possible, what is the best way to provide a custom config for this `proxy_cache` plugin?
Regards.

Baoyuantop:
You can configure a custom `proxy_cache` zone through `apisix.nginx.configurationSnippet` in `values.yaml` (lines 454-466). Use the `httpStart` section to add NGINX http-level directives:

```yaml
apisix:
  nginx:
    configurationSnippet:
      httpStart: |
        proxy_cache_path /tmp/disk_cache levels=1:2 keys_zone=my_cache:50m inactive=1d max_size=1g;
```

Alternatively, you can use `apisix.fullCustomConfig` (values.yaml lines 335-340) to provide a complete custom `config.yaml` with your desired NGINX configuration.

Closing as this is a usage question. If you still need help, please feel free to reopen.

PR #802, opened 2025-01-30
feat: Allow proxy_cache zones to be configured from values.yaml

brmejia:
This PR add missing functionality reported here [#667](https://github.com/apache/apisix-helm-chart/issues/667)

Additional proxy_cache zones can be defined using values.yaml like this:
```values.yaml
apisix:
  proxy_cache:
    zones:
      - name: memory_cache
        memory_size: 256m
```


brmejia:
Could someone give feedback or review this PR?
Thanks

oksman:
This would be a very nice feature! Currently it seems we need to overwrite the whole `config.yaml` which isn't that nice.

Baoyuantop (commented):
Thanks for working on this feature. The direction is useful because it turns the #667 workaround into an explicit values.yaml entry.

I do not think this is ready to merge yet. There are two points that should be addressed first:

1. The new schema is added as a top-level `proxy_cache`, while the actual template and docs use `apisix.proxy_cache.zones`. That means the schema does not validate the values path users will actually configure. Please either move the schema to the real values path or explain the current schema structure and add validation evidence.

2. This PR changes APISIX startup config rendering, but I do not see a `helm template`/chart-testing case showing a custom `proxy_cache.zones` entry rendered correctly. Please add a minimal rendering verification covering `name`/`memory_size` and the optional disk/cache_levels fields.

Also, the `toYaml -> fromYaml -> len` check in the template is not very clear for a list value. Using `with` or checking the list length directly would be easier to maintain.

PR #813, opened 2025-03-04
feat: add PROXY protocol support 

### LINKED ISSUE #812: Feature Request: Add Proxy Protocol Configuration Support to APISIX Helm Chart
I would like to add support for configuring the PROXY protocol in the APISIX helm chart. Currently, the PROXY protocol configuration exists in the template as commented code, but there's no way to enable or configure it via values.yaml.

 Motivation

The PROXY protocol allows load balancers to pass client connection information to APISIX. This is especially useful in Kubernetes environments where APISIX is deployed behind a load balancer, and the original client IP needs to be preserved.

Implementation Plan

- [ ]  Added proxy protocol configuration options to values.yaml with default values:
  proxyProtocol:
    enabled: false
    listenHTTPPort: 9181
    listenHTTPSPort: 9182
    enableTCPPP: false
    enableTCPPPToUpstream: false
- [ ] Updated configmap.yaml to conditionally include proxy_protocol configuration when enabled
- [ ] Added schema validation in values.schema.json for the new options
- [ ] Updated README.md with documentation for the new configuration options

These changes allow users to easily enable and configure proxy protocol support through the helm chart values, rather than having to manually modify the configmap.

casperakos:
I would like to add support for configuring the PROXY protocol in the APISIX helm chart. Currently, the PROXY protocol configuration exists in the template as commented code, but there's no way to enable or configure it via values.yaml.

**Motivation**

The PROXY protocol allows load balancers to pass client connection information to APISIX. This is especially useful in Kubernetes environments where APISIX is deployed behind a load balancer, and the original client IP needs to be preserved.

**Implementation**

- [x] Added proxy protocol configuration options to values.yaml with default values:

  proxyProtocol:
    enabled: false
    listenHTTPPort: 9181
    listenHTTPSPort: 9182
    enableTCPPP: false
    enableTCPPPToUpstream: false

- [x] Updated configmap.yaml to conditionally include proxy_protocol configuration when enabled

- [x] Added schema validation in values.schema.json for the new options

- [x] Updated README.md with documentation for the new configuration options

These changes allow users to easily enable and configure proxy protocol support through the helm chart values, rather than having to manually modify the configmap.

Fixes #812

### DIFF
--- a/charts/apisix/templates/configmap.yaml
+++ b/charts/apisix/templates/configmap.yaml
@@ -76,15 +76,15 @@ data:
       enable_http2: {{ .Values.apisix.enableHTTP2 }}
       enable_server_tokens: {{ .Values.apisix.enableServerTokens }} # Whether the APISIX version number should be shown in Server header
 
-      {{- if .Values.proxyProtocol.enabled }}
+      {{- if .Values.apisix.proxyProtocol.enabled }}
       proxy_protocol:                     # Proxy Protocol configuration
-        listen_http_port: {{ .Values.proxyProtocol.listenHTTPPort }}          # The port with proxy protocol for http, it differs from node_listen and admin_listen.
+        listen_http_port: {{ .Values.apisix.proxyProtocol.listenHTTPPort }}          # The port with proxy protocol for http, it differs from node_listen and admin_listen.
                                         # This port can only receive http request with proxy protocol, but node_listen & admin_listen
                                         # can only receive http request. If you enable proxy protocol, you must use this port to
                                         # receive http request with proxy protocol
-        listen_https_port: {{ .Values.proxyProtocol.listenHTTPSPort }}        # The port with proxy protocol for https
-        enable_tcp_pp: {{ .Values.proxyProtocol.enableTCPPP }}             # Enable the proxy protocol for tcp proxy, it works for stream_proxy.tcp option
-        enable_tcp_pp_to_upstream: {{ .Values.proxyProtocol.enableTCPPPToUpstream }} # Enables the proxy protocol to the upstream server
+        listen_https_port: {{ .Values.apisix.proxyProtocol.listenHTTPSPort }}        # The port with proxy protocol for https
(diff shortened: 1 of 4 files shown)


charts/apisix/templates/configmap.yaml
```diff
@@ -76,14 +76,16 @@ data:
...
-      #   enable_tcp_pp_to_upstream: true # Enables the proxy protocol to the upstream server
+      {{- if .Values.apisix.proxyProtocol.enabled }}
+      proxy_protocol:                     # Proxy Protocol configuration
+        listen_http_port: {{ .Values.apisix.proxyProtocol.listenHTTPPort }}          # The port with proxy protocol for http, it differs from node_listen and admin_listen.
+                                        # This port can only receive http request with proxy protocol, but node_listen & admin_listen
+                                        # can only receive http request. If you enable proxy protocol, you must use this port to
+                                        # receive http request with proxy protocol
+        listen_https_port: {{ .Values.apisix.proxyProtocol.listenHTTPSPort }}        # The port with proxy protocol for https
+        enable_tcp_pp: {{ .Values.apisix.proxyProtocol.enableTCPPP }}             # Enable the proxy protocol for tcp proxy, it works for stream_proxy.tcp option
+        enable_tcp_pp_to_upstream: {{ .Values.apisix.proxyProtocol.enableTCPPPToUpstream }} # Enables the proxy protocol to the upstream server
```
nic-6443, line 87: The proxy protocol ports (9181/9182) are configured in APISIX config here, but they're never exposed in the deployment template's `containerPorts` or in the `Service` spec. This means even with proxy protocol enabled, the ports won't be reachable from outside the pod.

You'd need to also update `deployment.yaml` to conditionally add these as container ports when `proxyProtocol.enabled` is true, and update the service template to expose them. Without that, a load balancer can't actually send PROXY protocol traffic to APISIX on these ports.

nic-6443 (changes requested):

PR #822, opened 2025-05-07
feat: support setting default http lua_shared_dict

dennispan:
This addresses the need from https://github.com/apache/apisix-helm-chart/issues/769

### DIFF
--- a/charts/apisix/templates/configmap.yaml
+++ b/charts/apisix/templates/configmap.yaml
@@ -206,6 +206,12 @@ data:
           {{ $dict.name }}: {{ $dict.size }}
         {{- end }}
         {{- end }}
+        {{- if .Values.apisix.nginx.luaSharedDicts }}
+        lua_shared_dict:
+        {{- range $dict := .Values.apisix.nginx.luaSharedDicts }}
+          - {{ $dict.name }}: {{ $dict.size }}
+        {{- end }}
+        {{- end }}
       {{- if .Values.apisix.nginx.configurationSnippet.main }}
       main_configuration_snippet: {{- toYaml .Values.apisix.nginx.configurationSnippet.main | indent 6 }}
       {{- end }}
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -440,13 +440,19 @@ apisix:
 
       stream: |
 
-    # -- Add custom [lua_shared_dict](https://github.com/openresty/lua-nginx-module#toc88) settings,
+    # -- Add custom [lua_shared_dict](https://github.com/openresty/lua-nginx-module?tab=readme-ov-file#lua_shared_dict) settings,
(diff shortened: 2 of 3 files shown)


charts/apisix/values.yaml
```diff
@@ -440,14 +440,20 @@ apisix:
...
       # - name: foo
       #   size: 10k
       # - name: bar
       #   size: 1m
 
+    # -- Override default [lua_shared_dict](https://github.com/apache/apisix/blob/master/conf/config.yaml.example#L250-L276) settings,
+    # click [here](https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/values.yaml#L27-L30) to learn the format of a shared dict
+    luaSharedDicts: []
+      # - name: prometheus-metrics
+      #   size: 20m
```
bzp2010, line 455: This example is inappropriate, prometheus is defined using the special command for `meta { lua_shared_dict }`, which cannot be overridden by `http { lua_shared_dict }`, but it does override other dict values, such as those used by limit-count.

ref: https://github.com/apache/apisix/blob/master/conf/config.yaml.example#L165-L168

This is a special case for creating shared memory between the HTTP subsystem and the stream subsystem.

bzp2010 (commented):
This solves problems partially, but does not include the `prometheus-metrics` mentioned in the comment

charts/apisix/values.yaml
dennispan: @bzp2010 I actually tried `meta` but it doesn't work. And overriding it under `http` does

At least it didn't work for me. Maybe I'm missing something.

Baoyuantop: @dennispan, can you provide your test steps? Maybe we can find the problem from them

dennispan: I used this to test (after a re-deployment to pick up the configuration changes):

```
/usr/local/openresty/nginx/sbin/nginx -T \
  -p /usr/local/apisix/ \
  -c conf/nginx.conf \
| grep lua_shared_dict
```

dennispan: @Baoyuantop @bzp2010 any update on this? Anything else needed from me to move this forward?

nic-6443 (approved):

charts/apisix/templates/configmap.yaml
swapnil-omni, line 1: This does not work.
 The nginx config is like this

 ```
lua_shared_dict:
      prometheus-metrics: 20m
```

 This created config like this

```
lua_shared_dict:
      - prometheus-metrics: 20m
```
  so this should be 

  ```
{{- range $dict := .Values.apisix.nginx.luaSharedDicts }}
           {{ $dict.name }}: {{ $dict.size }}
```

https://github.com/apache/apisix/blob/master/conf/config.yaml.example#L250-L276

PR #824, opened 2025-05-16
feat: add apisix.standaloneConfig.config

FoseFx:
Previously, a static `apisix.yaml` config map was created in standalone mode, with no way of updating it in the chart. now there is.


jompu:
Duplicate of https://github.com/apache/apisix-helm-chart/pull/717

PR #832, opened 2025-06-10
fix: switch to dict instead of list for nginx lua_shared_dict

dennispan:
Fixes issue from https://github.com/apache/apisix-helm-chart/pull/822


janiussyafiq:
duplicate #907

PR #844, opened 2025-07-07
Update Bitnami Etcd chart to 12.0.7

### LINKED ISSUE #788: etcd chart is old (9.7.3) and needs an update
It would be good to bump up version of etcd chart which is 9.7.3 while newest bitnami upstream is 9.15.2 or even 10.5.x

AndreaVida:
Fixes https://github.com/apache/apisix-helm-chart/issues/788


Baoyuantop:
It has now been updated to version 12.0.18

PR #852, opened 2025-07-25
chore: remove dashboard and updated docker image

juzhiyuan:
- [x] Remove Dashboard Docker Image from apisix chart
- [x] Update apisix/apisix docker image to Ubuntu based

PR #855, opened 2025-07-29
chore: update helm repo url

juzhiyuan:
Resolve https://github.com/apache/apisix-helm-chart/pull/852#issuecomment-3116839248


bzp2010 (approved):
While I approve it, I must point out: even if you modify the latest chart, you can't stop resolving that domain because the old helm in the index file of the helm repo is still using that old address, and unresolving it will cause people using the old helm chart to encounter errors. ref: https://github.com/apache/apisix-helm-chart/blob/gh-pages/index.yaml

I'm afraid you'll need to modify or rebuild the repository index if you want to fix this. Or maybe we can switch to Helm OCI mode.

PR #868, opened 2025-08-29
chore: upgrade etcd chart to latest version and use latest image

nic-6443:


nic-6443:
> Why do we need to use latest? Different etcd chart versions are tied to specific etcd application versions, and an older APISIX chart may not be fully compatible with the newest etcd release. Using latest also introduces unpredictability in deployments.

@AlinsRan Due to Bitnami's adjustment of the open-source strategy for Docker images: https://github.com/bitnami/containers/issues/83267, non-latest images will be unavailable for a period of time in the future.
> Bitnami will begin deprecating support for non-hardened, Debian-based software images in its free tier and will gradually remove non-latest tags from the public catalog. As a result, community users will have access to a reduced number of hardened images. These images are published only under the “latest” tag and are intended for development purposes

Therefore, unless we replace the etcd image with a historical image from the docker.io/bitnamilegacy repository, we can only use the latest image. Considering that our current requirement for etcd in the apisix helm chart is for quickly experiencing apisix and is only used for development and testing purposes without providing production guarantees for the etcd helm chart itself, we ultimately chose to use the latest image.

AlinsRan (approved):

membphis (approved):

PR #415, opened 2022-12-14
chore: use helm docs to generate documentation

shreemaan-abhishek:
Here's a brief description of what I did:

- prefixed the comments in the `values.yaml` file with `--` so that it could be used by the `helm-docs` tool for the description of fields to generate `README.md`.
- added some more comments as description to some yaml fields.
- added `.md.gotmpl` files which will be used to generate the `README.md` files.
- added github action step for checking if helm-docs was run by the user after making changes to the helm chart. This is to ensure that the README.md files are up to date.
- finally, I used the helm-docs tool to regenerate the README.md file.

Closes #334

PR #871, opened 2025-09-05
docs: add missing wasm parameters

kayx23:
Wasm's `priority`, `file`, and `http_request_phase` were previously documented but got removed in https://github.com/apache/apisix-helm-chart/pull/415 which switched to using `helm-docs` to generate the docs.

The missed configuration could lead to:

image

This PR adds these parameters back.

### DIFF
--- a/charts/apisix/values.yaml
+++ b/charts/apisix/values.yaml
@@ -554,9 +554,16 @@ apisix:
     cmd: ["/path/to/apisix-plugin-runner/runner", "run"]
 
   wasm:
-    # -- Enable Wasm Plugins. See [wasm plugin](https://apisix.apache.org/docs/apisix/next/wasm/)
+    # -- Enable Wasm Plugins. See [wasm plugin](https://apisix.apache.org/docs/apisix/next/wasm/).
     enabled: false
+    # -- Set Wasm plugin name.
     plugins: []
+    # -- Set wasm plugin priority.
+    priority: 7999
+    # -- Set path to Wasm plugin.
+    file: ""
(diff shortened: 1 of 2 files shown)


juzhiyuan (approved):

charts/apisix/values.yaml
```diff
@@ -554,9 +554,16 @@ apisix:
...
   wasm:
-    # -- Enable Wasm Plugins. See [wasm plugin](https://apisix.apache.org/docs/apisix/next/wasm/)
+    # -- Enable Wasm Plugins. See [wasm plugin](https://apisix.apache.org/docs/apisix/next/wasm/).
     enabled: false
+    # -- Set Wasm plugin name.
     plugins: []
+    # -- Set wasm plugin priority.
+    priority: 7999
+    # -- Set path to Wasm plugin.
+    file: ""
```
nic-6443: `priority`, `file`, and `http_request_phase` are per-plugin properties in APISIX's config, not top-level wasm settings. The existing configmap template just does `toYaml .Values.apisix.wasm.plugins`, so these top-level values won't be used in the generated config at all.

The correct structure should have these as fields within each plugin entry:

```yaml
wasm:
  enabled: false
  plugins:
    - name: "example-plugin"
      priority: 7999
      file: "/path/to/plugin.wasm"
      http_request_phase: "rewrite"
```

As-is, setting `apisix.wasm.priority` or `apisix.wasm.file` would have no effect on the rendered config. The docs should show the correct per-plugin structure instead.

nic-6443 (changes requested):

PR #884, opened 2025-10-16
chore: update crd for ingress

AlinsRan:
ref: https://github.com/apache/apisix-ingress-controller/tree/master/config/crd/bases


ronething:
charts/apisix-ingress-controller/crds/gwapi-crds.yaml Do we also need to update this file?

nic-6443 (approved):

PR #896, opened 2025-10-28
Address the issue of configuration files being unable to update in standalone mode due to being a ConfigMap

storezhang:


Baoyuantop (commented):
Thanks for the contribution. I do not think this PR is mergeable in its current form because the diff does not fix the root cause of the standalone ConfigMap update issue, and it also introduces several unrelated high-risk changes.

First, the `deployment.yaml` change copies `/apisix-config/apisix.yaml` into `{{ .Values.apisix.configDir }}` once during container startup and then symlinks APISIX to that copied file. ConfigMap volume updates still happen under the mounted `/apisix-config` directory, and they will not be synchronized into the copied regular file. So if the goal is hot reload for APISIX standalone mode, this change can make the situation worse by making APISIX read a startup-time copy instead of the projected ConfigMap file. If the goal is to apply changes via a rolling restart, the chart needs a pod-template checksum or another restart mechanism, which this PR does not add.

Second, `configmap.yaml` changes `proxy_protocol` from commented sample configuration into rendered default configuration and references `.Values.apisix.proxyProtocol.*`, but `values.yaml` does not define the corresponding default structure. This is unrelated to the standalone ConfigMap update issue and may break Helm rendering or change the default APISIX configuration for all deployment modes.

Also, `.github/workflows/cos.yaml` adds a Tencent COS publishing workflow. That is unrelated to this PR's stated goal and introduces an external action, cloud credentials, and push-triggered publishing behavior. Please remove it from this PR.

Please narrow this PR to the standalone ConfigMap update issue only, clarify whether the expected behavior is hot reload or rolling restart, and add minimal reproduction plus verification, such as `helm template` output and APISIX logs or behavior after updating the ConfigMap. The proxy protocol and COS workflow changes should be removed or split into separate PRs.

PR #899, opened 2025-11-03
feat: bump apisix-ingress-controller's adc to 0.22.1

bzp2010:
In the currently used ADC 0.21.x, the ADC running in server mode suffers from a connection leak issue. This manifests as the ADC establishing an increasing number of connections over time, which can only be terminated by the server actively closing them.

ADC 0.22.x has addressed this issue by limiting the connection cap per host to 256 and proactively releasing connections after 60 seconds of inactivity. This significantly improves the previous connection leak problem.

This PR upgrades the ADC version for AIC.

### DIFF
--- a/charts/apisix-ingress-controller/Chart.yaml
+++ b/charts/apisix-ingress-controller/Chart.yaml
@@ -24,7 +24,7 @@ keywords:
   - nginx
   - crd
 type: application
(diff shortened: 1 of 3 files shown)


ronething (approved):
lgtm

PR #904, opened 2025-11-14
feat: add configurable real_ip_header option for nginx

vichaos:
### Summary
This PR makes the nginx `real_ip_header` directive configurable via `values.yaml`, allowing users to customize which header is used to extract the real client IP address.

### Changes
- Added `apisix.nginx.realIpHeader` configuration option in `values.yaml` with a default value of `"X-Real-IP"`
- Updated `configmap.yaml` template to use the configurable value instead of hardcoded `"X-Real-IP"`
- Maintains backward compatibility with existing deployments (default remains `"X-Real-IP"`)

### Why This Change is Necessary

Different deployment environments and proxy setups use different headers to pass the original client IP address:

1. **Load Balancer Variations**: Different cloud providers and load balancers use different headers:
   - AWS ALB/ELB often uses `X-Forwarded-For`
   - Cloudflare uses `CF-Connecting-IP`
   - Google Cloud Load Balancer may use `X-Forwarded-For`
   - Some proxies use `X-Real-IP` (current default)

2. **Multi-Layer Proxy Environments**: In complex deployments with multiple proxy layers (e.g., CDN → Load Balancer → Ingress → APISIX), the appropriate header may differ based on architecture.

3. **Security and Rate Limiting**: Correctly identifying client IPs is crucial for:
   - Accurate rate limiting based on client IP
   - Security plugins that rely on IP-based access control
   - Logging and analytics
   - Compliance requirements

4. **Flexibility**: Without this configuration option, users would need to either:
   - Fork the chart and modify it
   - Use post-deployment patches
   - Accept incorrect IP detection in their environment

This change provides the flexibility needed for various deployment scenarios while maintaining sensible defaults.

### Usage Example

```yaml
apisix:
  nginx:
    realIpHeader: "X-Forwarded-For"  # For AWS ALB/ELB environments
```

Or via Helm CLI:
```bash
helm install apisix . --set apisix.nginx.realIpHeader="CF-Connecting-IP"
```

### DIFF
--- a/charts/apisix/README.md
+++ b/charts/apisix/README.md
@@ -105,6 +105,8 @@ The command removes all the Kubernetes components associated with the chart and 
 | apisix.nginx.logs.errorLog | string | `"/dev/stderr"` | Error log path |
 | apisix.nginx.logs.errorLogLevel | string | `"warn"` | Error log level |
 | apisix.nginx.luaSharedDicts | list | `[]` | Override default [lua_shared_dict](https://github.com/apache/apisix/blob/master/conf/config.yaml.example#L250-L276) settings, click [here](https://github.com/apache/apisix-helm-chart/blob/master/charts/apisix/values.yaml#L27-L30) to learn the format of a shared dict |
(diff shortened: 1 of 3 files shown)


ngclinflows (commented):
Don't forget to run the `make helm-docs` command to update the documentation.

PR #919, opened 2025-12-16
feat: add separate chart for CRDs

olegbuevich:
# Description

This PR adds a dedicated chart for CRDs.

# Problem

- Issues with multitenant cluster deployments
- Conflicts when the Gateway API CRDs already exist

# Solution

- Separate CRDs chart
- The Gateway API CRDs installation is optional
- Possibility to select Gateway API channel to install


gw-kamara:
Just throwing my two cents in - This is really a needed feature

finnegan-haack-lhind:
Any chance this PR might get merged? Without this change, deploying the ingress controller on any GKE cluster using the GatewayAPI is not possible, as experimental CRDs are not allowed. Trying to install the ingress controller helm chart leads to:

```
customresourcedefinitions.apiextensions.k8s.io "grpcroutes.gateway.networking.k8s.io" is forbidden: ValidatingAdmissionPolicy 'enforce-gateway-standard-channel' with binding 'enforce-gateway-standard-channel-binding' denied request: All Gateway API CRDs must belong to the 'standard' channel. Experimental CRDs are not permitted.
```

Skipping the CRDs for the helm chart, using the GKE provided GatewayAPI CRDs and deploying the APISIX ingress controller specific CRDs using

```
kubectl apply --server-side -f https://raw.githubusercontent.com/apache/apisix-helm-chart/master/charts/apisix-ingress-controller/crds/apisixic-crds.yaml
```

also does not work, as at least the GRPCRoute CRD is not supplied by GKE and causes a crash loop in the ingress controller.

juzhiyuan:
Thanks for reporting, I may request my team to review, this takes a few days.

janiussyafiq:
#1010 would support this request

PR #916, opened 2025-12-15
 feat(ingress-controller): add imagePullSecrets support with global fallback

ArnaudCHT:
## What
Add deployment.imagePullSecrets support in ingress-controller with fallback to global.imagePullSecrets.

## Why
The ingress-controller deployment had no way to configure imagePullSecrets, forcing users to manually patch deployments or fork the chart.

## Changes

- Add deployment.imagePullSecrets configuration option
- Falls back to global.imagePullSecrets if not set locally
- Backward compatible and works standalone

https://github.com/apache/apisix-helm-chart/issues/883

## Testing

To test this PR, you can use the following commands:
```bash
# Test with global.imagePullSecrets
helm template . -s templates/deployment.yaml \
  --set 'global.imagePullSecrets[0].name=global-registry-secret'

# Test with deployment.imagePullSecrets (local override)
helm template . -s templates/deployment.yaml \
  --set 'deployment.imagePullSecrets[0].name=my-registry-secret'

# Test priority: deployment takes precedence over global
helm template . -s templates/deployment.yaml \
  --set 'global.imagePullSecrets[0].name=global-registry-secret' \
  --set 'deployment.imagePullSecrets[0].name=my-registry-secret'

# Test standalone (no imagePullSecrets)
helm template . -s templates/deployment.yaml
```

PR #948, opened 2026-03-03
charts/ingress-controller: add global.imagePullSecrets support

martin-schulze-e2m:
### Motivation

Pulling the ingress-controller from a registry with authentication is currently not possible. This makes working around the Dockerhub ratelimiting hard.

### Details

Since the ingress-controller is used as a subchart by apisix, which already defines `global.imagePullSecrets`, I decided to reuse them.


martin-schulze-e2m:
I just found https://github.com/apache/apisix-helm-chart/pull/916, it is beyond my why me previous search did not turn that up.

PR #962, opened 2026-04-15
chore: upgrade apisix-ingress-controller from v1.1.2 to v1.2.0 and bump apisix chart to support it

Guerlielton:
This PR upgrades the apisix-ingress-controller dependency to version 1.1.2.

The main purpose of this upgrade is to add support for imagePullSecrets, enabling the use of private container registries.

This feature was introduced in PR #916 , but the apisix chart dependency was not updated at that time. This change ensures the dependency version aligns with the introduced functionality.

No breaking changes are expected, and this update aligns the deployment with current requirements for authenticated image pulls.


Guerlielton:
Hi @Baoyuantop and @AlinsRan I'm in doubt if I need to bump one minor version on apisix-2.14.0 like point to apisix-2.14.1

martin-schulze-e2m:
> The main purpose of this upgrade is to add support for imagePullSecrets, enabling the use of private container registries.

AFAIK the imagePullSecrets fix is not yet packaged in a release, so we would need a new AIC chart version (1.1.3?) first.

Guerlielton:
Hi @martin-schulze-e2m, this change is already there after merging this #916 but the version stays 1.1.2 to follow the right way. Yes, we need to put a new version from AIC I'll do this and wait for the team review.

Guerlielton:
Hi, as @martin-schulze-e2m  mentioned, I’ve opened a new PR (#969) to update the apisix-ingress-controller to version 1.1.3.

Once that PR is merged, I’ll update the Chart.lock in this PR to reflect the new 1.1.3 version.

martin-schulze-e2m:
Probably will be superseeded by #973

Guerlielton:
Thanks @martin-schulze-e2m and team I'll close this PR.

PR #1003, opened 2026-08-26
feat(apisix): render extraContainers through tpl

ahmedgabers:
extraContainers entries are raw container specs, so an image reference inside them cannot be composed from other values the way first-class image fields can. Rendering the list through `tpl` lets entries carry template expressions (e.g. `image: "{{ .Values.someRepository }}:{{ .Values.someTag }}"`), matching the common pattern for extra-container passthroughs. Plain specs render unchanged.

Verified with `helm template`: a templated entry expands correctly, and untemplated values render byte-identical to the current output. Chart version bumped 2.17.0 -> 2.17.1; values doc comment and README row updated.


ahmedgabers:
Fixes #1005.

PR #1004, opened 2026-08-26
feat(ingress-controller): support annotations on the webhook configuration

ahmedgabers:
The ValidatingWebhookConfiguration renders with no annotations, so external CA managers cannot attach to it. With `webhook.certificate.provided`, operators supply a serving certificate (commonly a cert-manager Certificate) but must hand-copy the CA into `webhook.certificate.caBundle` and re-copy it whenever it rotates.

An `webhook.annotations` knob lets cert-manager's cainjector own the caBundle instead (`cert-manager.io/inject-ca-from: <namespace>/<certificate>`) — the standard pattern for webhook trust management.

Verified with `helm template`: the annotation renders onto the webhook configuration when set; unset output is byte-identical to today. Chart version bumped 1.3.0 -> 1.3.1; values doc comment and README row updated.

### DIFF
--- a/charts/apisix-ingress-controller/Chart.yaml
+++ b/charts/apisix-ingress-controller/Chart.yaml
@@ -24,7 +24,7 @@ keywords:
   - nginx
   - crd
 type: application
-version: 1.3.0
+version: 1.3.1
 appVersion: 2.2.0
 sources:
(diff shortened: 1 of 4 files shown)


ahmedgabers:
Fixes #1006.

